PR #10, opened 2020-05-27
Bootstrap React App - Ready for Review!

ginonott:
Scaffolding for the client side application. The bulk of the LOC comes from the package-lock.

# State Management
I chose [Redux](https://redux.js.org/) because its React agnostic so it can be used with other pieces of our application that may not directly communicate with React as well as the fact that its tried and true in the React community.

# Component Library
See #12 for discussion on choosing a component library. It will not be a part of this PR.

# Clientside Routing
[https://reacttraining.com/react-router/web/guides/quick-start](React Router) seems to be the most popular and widely used.

# Folder / Code Structure
There are a couple common ways to structure code in frontend repos. One way is to have a folder for components, reducers, utilities, etc. Another is to group by feature / domain. I propose that we go with the latter especially as the React UI may be very separate from the Blockly and Simulator code. Folders may have sub folders for specific features within these main features.

The top level `src` folder will also contain the entry point of the application as well as some files that are global to all features such as the main application, the redux store, etc.

# Formatting
Formatting is done with prettier. Prettier formats the code the way prettier wants. This has been extremely helpful in keeping the code base consistent and clean and preventing discussions about how code should aesthetically look. See https://prettier.io/. It should run on each pre-commit via husky (https://github.com/typicode/husky).

# CI
To ensure stability, I think before merging this we should setup a CI solution using either github actions, circle ci, travis ci or something else. Github actions might be the quickest as its built into github. The solution needs to call:
```
npm run build
npm run test
```

See See https://github.com/ginonott/Simulator/runs/716718533?check_suite_focus=true for an example of the workflow provided in this PR.

# TODO:
- [x] Bootstrap react app (8f8a6e1)
- [x] Setup Redux (6dac411)
- [x] Setup React Redux (c727e0d)
- [x] Setup React Router (3b0a70a)
- [x] ~Pick a UI Component Library - input needed!~ See #12 
- [x] Pre-commit checks (formatter and lint) (1b9fc68)
- [x] Setup a github project CI to build and test app (3e9e411)


JamesWP:
Wow this is looking great.

README.old.md
```diff
@@ -0,0 +1 @@
+# Simulator
```
JamesWP, line 1: this can be removed

package.json
```diff
@@ -0,0 +1,65 @@
...
+    "checkFormat": "prettier --check src",
+    "eject": "react-scripts eject"
+  },
+  "eslintConfig": {
+    "extends": "react-app"
+  },
+  "browserslist": {
+    "production": [
+      ">0.2%",
+      "not dead",
```
JamesWP, line 43: love this part :D

src/store.ts
```diff
@@ -0,0 +1,16 @@
+import { configureStore, combineReducers } from "@reduxjs/toolkit";
+import { userReducer } from "./User/userReducer";
+
+// Type Safe Reducers - se
+const rootReducer = combineReducers({
+  user: userReducer.reducer,
```
JamesWP, line 6: do we need to add the other reducers in here?

ginonott: Oops - good catch! Yep. This where each reducer would need to go. I also see I mysteriously stopped typing the comment. I will fix the comment and add the other two reducers

PR #13, opened 2020-05-29
add basic layout

JamesWP:


src/AppRouter.tsx
```diff
@@ -11,9 +11,14 @@ export const AppRouter = () => {
   return (
     <Switch>
       <Route path="/" exact>
-        {/* TODO: Break into a view component*/}
-        <Blockly />
-        <RobotSimulator />
+        <div className="grid-container">
```
ginonott, line 14: I think this is a good time to move this to its own component such as "SimulatorView" or something along those lines. That component can then import the above mentioned CSS file.

src/AppRouter.tsx
```diff
@@ -11,9 +11,14 @@ export const AppRouter = () => {
...
+          <div className="menu-bar"></div>
```
ginonott, line 15: Since the menu bar is usually static on website (always present on the top), I think we should move this to the `App` component so that it lives alongside AppRouter. So it would look like:
```
App
  Header
  Router
    Simulator & subcomponents
  Footer
```
This will prevent the header and footer from re-rendering / needing to be imported into every page in the future.

src/AppRouter.tsx
```diff
@@ -11,9 +11,14 @@ export const AppRouter = () => {
...
+          {/* TODO: Break into a view component*/}
+          <Blockly />
+          <RobotSimulator />
+          <div className="control-panel"></div>
+          <div className="footer"></div>
```
ginonott, line 20: Same as above comment.

src/App.css
```diff
@@ -0,0 +1,44 @@
...
+  grid-template-rows: 50px 30vw 1fr 30px;
+  gap: 1px 1px;
+  grid-template-areas:
+    "Menu-Bar Menu-Bar Menu-Bar       Menu-Bar"
+    ".        Blockly  Simulator      ."
+    ".        Blockly  Control-Panel  ."
+    "Footer   Footer   Footer         Footer";
+}
+
+.blockly-ui {
```
ginonott, line 22: Can you move this block to "Blockly.css" file and import it in the Blockly file?

src/App.css
```diff
@@ -0,0 +1,44 @@
...
+  grid-area: Blockly;
+  background-color: #e8f5e9;
+}
+
+.robot-simulator {
```
ginonott, line 27: Can you move this to "RobotSimulator.css" file and import it in the RobotSimulator file?

src/App.css
```diff
@@ -0,0 +1,44 @@
+html,
+body,
+#root,
+.App,
+.grid-container {
+  height: 100%;
+  margin: 0;
+}
+
+.grid-container {
```
ginonott, line 10: Cool use of CSS grid!

src/App.css
```diff
@@ -0,0 +1,44 @@
...
+  display: grid;
+  grid-template-columns: 30px 1fr 1fr 30px;
+  grid-template-rows: 50px 30vw 1fr 30px;
+  gap: 1px 1px;
+  grid-template-areas:
```
ginonott, line 15: 👍

JamesWP:
ready for re-review @ginonott

src/SimulatorView/SimulatorView.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { FunctionComponent } from "react";
```
ginonott, line 1: From the gitter chat:
Only possible suggestion would be to make the "SimulatorView" folder more generic - maybe something like "Layout" or "Views"? Totally up to you.

PR #17, opened 2020-05-31
Added framework for using three js in project

JamesWP:
The integration is done in two parts, one react component, and one
typescript class. This seperates the react code from the 3D code.


src/RobotSimulator/RobotSimulator.tsx
```diff
@@ -1,6 +1,68 @@
...
+
+  private onresizeCb = this.onresize.bind(this);
+
+  private sim: Sim3D | null = null;
+  private isRendering = false;
+
+  componentDidMount() {
+    this.onresize();
+
+    window.addEventListener("resize", this.onresizeCb);
```
ginonott, line 19: Should we throttle or debounce this? Is resizing an expansive operation? See https://css-tricks.com/debouncing-throttling-explained-examples/.

TLDR: We can use `lodash/throttle` or `lodash/debounce` to only call the resize once after the user stops resizing the window or only X amount of times per second.

This can also be a follow up PR.

JamesWP: lets make it a follow up, but i agree too

PR #18, opened 2020-06-02
Blockly integration

Casper-Oakley:
This PR contains a full skeleton for the default blockly widget used in the blockly-interaction-example repo. The rendering is included, though there are no redux-integration elements involved yet.

I have ported over the block highlighting code and code generation functions, though they don't have a react part yet.


src/BlocklyInterface/BlocklyInstance.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  constructor(workspaceArea: HTMLDivElement, toolbox: HTMLElement) {
+    this.workspace = Blockly.inject(workspaceArea, { toolbox });
+
+    this.setupInterpretation();
+    this.resizeBlockly();
+  }
+
+  setupInterpretation() {
+    const generator = (Blockly as any).JavaScript as BlocklyJavaScript;
```
Casper-Oakley, line 25: This is of course the wrong way to do this, but looks like the way Blockly generators are imported is broken with their typescript definitions (i.e. importing of the specified generator after Blockly is imported doesn't change the type definitions). I tried merging in declarations but got nowhere.

Anyone have any good ideas?

ginonott: Is the issue that Blockly is missing types so you are trying to type Blockly? If so, you can make a .d.ts file that defines blockly and its types.

Casper-Oakley: It's slightly more tricky than that. Blocky defines a .d.ts, but it doesn't include it for the generators (as they aren't necessarily all used). In the second blocky import at the top of the file, the import adds the JavaScript key to the Blocky module, but doesn't do any types.

ginonott: Can we have a single function that returns the typed generator?

src/BlocklyInterface/BlocklyEditor.tsx
```diff
@@ -0,0 +1,74 @@
...
+    workspaceStyle.width = wrapperRef.current.offsetWidth + "px";
+    workspaceStyle.height = wrapperRef.current.offsetHeight + "px";
+  }
+
+  // Initialize blockly and return destruction callback
+  useEffect(() => {
+    resizeBlocklyRegion();
+
+    blocklyRef.current = new BlocklyInstance(
+      workspaceAreaRef.current!,
```
ginonott, line 51: Is it safe to make this assumption?

src/BlocklyInterface/BlocklyEditor.tsx
```diff
@@ -0,0 +1,74 @@
...
+      toolbox.current!
```
ginonott, line 52: Is it safe to make this assumption?

src/BlocklyInterface/BlocklyEditor.tsx
```diff
@@ -0,0 +1,74 @@
...
+  useEffect(() => {
+    function onResizeHandler() {
+      resizeBlocklyRegion();
+
+      blocklyRef.current?.resizeBlockly();
+    }
+
+    window.addEventListener("resize", onResizeHandler);
+
+    return window.removeEventListener.bind(window, "resize", onResizeHandler);
```
ginonott, line 66: 👍

src/BlocklyInterface/BlocklyInstance.ts
```diff
@@ -0,0 +1,47 @@
+import Blockly from "blockly";
+import "blockly/javascript";
+
+declare interface BlocklyJavaScript {
+  STATEMENT_PREFIX: string;
+  addReservedWords: any;
```
ginonott, line 6: Can you add proper types here?

src/BlocklyInterface/BlocklyInstance.ts
```diff
@@ -0,0 +1,47 @@
...
+  workspaceToCode: any;
```
ginonott, line 7: Can you add proper types here?

Issue #12, opened 2020-05-28
Pick Component UI Library


ginonott:
It'll be faster and provide a better user experience if we pick an open source component library. Here are some options https://blog.logrocket.com/top-10-react-component-libraries-for-2020/ to look into.

Issue #20, opened 2020-06-04
Create VM react component


JamesWP:
This should integrate the [js-interpreter npm](https://www.npmjs.com/package/js-interpreter) package and have a simple react UI where the state of the execution is displayed. 

The following buttons should be available:
* run code ( loads the latest code, and starts execution) (only available if the vm is stopped.)
* reset ( reset the vm, allows 'run code' to be clicked again )
* pause ( execution stops the code from executing )
* resume ( continue execution of code )

This component should subscribe to events from a redux store containing the latest code.

PR #23, opened 2020-06-05
Add JS Interpreter

ginonott:
This PR will connect blockly and the vm together via redux. ~I know this is pretty ugly looking. I'm thinking a follow up PR could be to add a UI library as per #12 and replace our divs with background-colors with real components.~ I went with Fluent UI. We can always change this later.

Play with it here: https://deploy-preview-23--fruk-simulator.netlify.app/


JamesWP:
Sounds good, will look later. This will help #20 too

src/BlocklyInterface/BlocklyEditor.tsx
```diff
@@ -45,12 +51,33 @@ export const BlocklyEditor: FunctionComponent<BlocklyEditorProps> = ({
 
   // Initialize blockly and return destruction callback
   useEffect(() => {
+    function handleBlocklyChange(event: BlocklyEvent) {
+      if (!blocklyRef.current) {
+        return;
+      }
+
+      console.log(event);
```
Casper-Oakley, line 59: can probs remove

src/JavascriptVM/vm.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AppDispatch } from "../store";
+import { blocklySlice } from "../BlocklyInterface/blocklySlice";
+
+export const getIntepreter = (code: string, dispatch: AppDispatch) => {
+  function initInterpreter(
+    intepreter: Interpreter,
+    globalObject: GlobalObject
+  ) {
+    const highlightBlock = intepreter.createNativeFunction((id: string) => {
+      dispatch(blocklySlice.actions.highlightBlock({ blockId: id }));
```
Casper-Oakley, line 11: in https://github.com/FRUK-Simulator/blockly-intepretation-example/blob/master/src/interpreterwrapper.ts#L12 , there's an extra bit of state to track when the previous highlightBlock call was fired. This is so the [step button](https://github.com/FRUK-Simulator/blockly-intepretation-example/blob/master/src/interpreterwrapper.ts#L29) will step to the next blockly step, as opposed to the next intepreter step, meaning that hitting step will highlight the next block. It'd be nice to have that here as well

ginonott: Great feedback! I was trying to figure out how to correspond the interpreter code with the blockly code! Didn't dawn on me to just wait until the next block was highlighted!

JamesWP:
@ginonott looks a million times better now :), but i still see blocks being highlighted after i press the stop button

ginonott:
> LGTM, though we should think about a solution to the infinite loop for another pr.

Agreed. There's just one more change I want to make before merging this. I want to decouple the JS VM from the controls to it.

JamesWP:
> > LGTM, though we should think about a solution to the infinite loop for another pr.
> 
> Agreed. There's just one more change I want to make before merging this. I want to decouple the JS VM from the controls to it.
> 
> Or I can make that a follow up PR.

i don't mind which, ill let you decide

Issue #29, opened 2020-06-10
Blockly should be disabled while the VM is running


Casper-Oakley:
As per title, users should be unable to modify their blockly program during execution.

Blockly itself has no way to switch into a read only mode after startup, but it [is recommended](https://groups.google.com/forum/#!topic/blockly/NCukwTKMR0U) to instead just cover the blockly editor with a div, de-select the current block and clear the blockly clipboard to in essence do the same.

Presumably after execution, the previously selected block, and the blockly clipboard will want to be restored.

PR #38, opened 2020-06-13
Block blockly modifications while VM running

Casper-Oakley:
Fix for #29 - blocks user interactions via the mouse, and drops current block selection. Restores block selection on finishing execution.

The presentation side for explaining to the user why blockly editing is disallowed probs needs a little more thought - does anyone have any good ideas for that? E.g. somewhere to put some text to state blockly is read-only while the VM is running


src/BlocklyInterface/BlocklyEditor.tsx
```diff
@@ -51,12 +63,20 @@ export const BlocklyEditor: FunctionComponent = () => {
...
       }
 
-      if (
-        event instanceof Events.BlockMove ||
-        event instanceof Events.BlockChange
-      ) {
+      dispatch(vmSlice.actions.setCode({ code: blocklyRef.current.getCode() }));
+    }
+
+    //function handleBlocklyUiEvent(event: Events.Ui) {
```
Casper-Oakley, line 69: Any typescript aficionados - the blockly provided types for Events.Ui doesn't for whatever reason contain the additional Events.Ui fields that aren't in the parent class, Events. They are however supported and in [the official docs](https://developers.google.com/blockly/guides/configure/web/events#blocklyeventsui).

Does anyone know if its possible to append to a class type def inside an external module?

ginonott:
> The presentation side for explaining to the user why blockly editing is disallowed probs needs a little more thought - does anyone have any good ideas for that

The common thing to do here would be to put a semi transparent gray overlay over the blockly area. When hovering over it, displaying https://developer.microsoft.com/en-us/fluentui#/controls/web/tooltip can provide an explanation.

JamesWP:
Do we need to store the previous selection? Can we just wipe it? The running code will have ran on from what was selected previously right?

Casper-Oakley:
> > The presentation side for explaining to the user why blockly editing is disallowed probs needs a little more thought - does anyone have any good ideas for that
> 
> The common thing to do here would be to put a semi transparent gray overlay (https://developer.microsoft.com/en-us/fluentui#/controls/web/overlay) over the blockly area. When hovering over it, displaying https://developer.microsoft.com/en-us/fluentui#/controls/web/tooltip can provide an explanation.

I actually thought of this while looking at the code, then decided against it as i didn't want to obscure the user's view of the blockly highlighted blocks - I'll put an overlay on and throw an image of it on here, and if we're happy with it code it in


> Do we need to store the previous selection? Can we just wipe it? The running code will have ran on from what was selected previously right?

We certainly could just wipe the selection for sure! Its just there so if a user has a block selected, then hits start and stop, that same block is selected - up to us if needed or not. I don't think it affects the running of the code, but I totally could be wrong

ginonott:
> decided against it as i didn't want to obscure the user's view of the blockly highlighted blocks 

That's a very good point. Maybe just to the tooltip then and make the cursor a disabled pointer? Css: `pointer: disabled`

Casper-Oakley:
I've removed the previousSelection logic entirely as it doesn't feel needed and i've added a tooltip. I tried to add the fluent UI tooltip, but it worked by optionally wrapping around the blockly workspace area, which meant react would re-render the workspace area if the tooltip was present or not.

This ended up un-initialising the blockly instance itself, and wiping the blockly data. Hence i've just used a title - lemme know if theres anything I can do about the fluent UI one as it looks a little nicer with one

PR #39, opened 2020-06-13
Integrate ControlHubEmulator

eurocoinsalbum:


ginonott:
Cool! This looks like a really good start  - I just have some broad questions.  I like the use of redux for this.

1. Does the ControlHubEmulator need to be its own package? Can it live in Simulator or SimulatorCore? It seems right now you have it doing 2 pieces: keeping track of motor / sensor state and providing a view to render the output. Maybe it makes sense to split these two?
2. Can we use regular HTML and React components to render the log and hub? The canvas api is really meant for rendering complex graphics like svgs, 2d games, and 3d web gl views. You lose a lot of benefits of the browser like being able to style stuff with CSS, accessibility, automatic resizing, etc.
3. It seems like the reducer you wrote stores the same information as the control hub object. Maybe it makes sense to integrate the control hub emulator into simulator core and then use redux to dispatch actions from blockly? The flow might look like this for setting motor output:
```
blockly -> code -> vm -> dispatches action (setMotorPower, etc) -> RobotSimulator (React Component) -> Robot (SimulatorCore with ControlHub)
```
And for setting input (touch sensors, etc)
```
Robot (SimulatorCore/ControlHub) -> RobotSimulator (React Component) ->dispatch action (setTouchInput) -> redux  -> vm
```

.vs/VSWorkspaceState.json
```diff
@@ -0,0 +1,16 @@
+{
```
ginonott, line 1: please add .vs to .gitignore

JamesWP:
First, thanks for putting in the pr! Looking forward to many future prs.

I broadly agree with @ginonott on his points. Though I’m not sure how integrated we want the state for robot things and redux. I know redux manages state, but I’m not sure we want things like setting motor power to trigger events.
‘’’
Vm -> robot.setmotorpower(id, 0.4) -> SimulatorCore.setrobotmotorpower(...) -> plank-js ...
‘’’
Having said that, If your component needs to store that value then it might be able to hook a redux event in somewhere and hook the value from there. 

The last way would be to periodically check what the latest values are by accessing the values from the simulator. ( as dumb as this is, I think it’s the most future proof ) 
Thoughts?

ginonott:
One reason why I think using redux as a middleman would be a good idea is that it provides a one directional source of truth for the state of the simulator that blocky, react, and the robot can read from that is easily debuggable. The redux debugger let's you see a stream of the global of events which is really nice.

eurocoinsalbum:
thanks for the feedback. For me the control hub is the truth and the central instance regarding the commands and reading the sensors like in real world. It makes sense to model it as near to reality as possible to stay future safe.
1. I will move the ControlHub into the SimulatorCore.
1b. The model(state) and UI are split already into two files but I can move them into 2 different directories.
2. I would love to use HTML but I get only a canvas...
3. I think the react component for ControlHub should be a follower of the ControlHubEmulator and not vice-versa. The UI should visualize the model and not the UI triggers the model.
I have no react knowledge but for me it looks like a react component is the holder of the view.
4. "accessing the values from the simulator" I think no one should query values from the Simulator except the controlHub which queries the sensors (distance, "GPS", touch, ...). Especially no one should query motor state from the Simulator because this is just the renderer (more or less). Even the simulator receives commands from the control hub.

zhiquanyeo:
I think that the ControlHub should be in Simulator really, since it is FTC specific. You can have the ControlHub be the one that interacts with SimulatorCore and issues "low level" commands to the virtual robot in SimulatorCore. I.e. think of SimulatorCore as the "hardware" that is platform (or environment) agnostic

eurocoinsalbum:
The controlHubEmulator is not meant to be FTC specific even if I use the name and the idea. For me it's something like an RC receiver which also can transmit sensor data back.
And this thing is hardware as well - like the robot. There is no robot without a receiver right?

zhiquanyeo:
We are going to be exposing a `RobotHandle` (or something similar) from the SimulatorCore side that allows manipulation of the virtual robot via methods like `setMotorPower(channel, value)` or `getDigitalSensorValue(channel)`. The last commit in https://github.com/FRUK-Simulator/SimulatorCore/pull/30 shows a very rough cut of this.

eurocoinsalbum:
OK

ginonott:
> I know redux manages state, but I’m not sure we want things like setting motor power to trigger events.

Do you have a reason for this?

PR #40, opened 2020-06-15
Integrating ControlHub

eurocoinsalbum:
I have no clue about react and redux ;-) so my implementation is a copy/paste/adopt of the existing implementation of "HighlightBlock" and the "Sim3D"-View. I can change anything but first someone with react/redux knowledge should then refactor the existing implementation of those so that I can make it similar.
In this PR you will find:
1. ControlHub which act as first point of contact for the blockly code: Blockly->VM->ControlHub ->All listeners
2. View for state of ControlHub (motor power 0-3) (listener of ControlHub)
3. View for log of ControlHub (timestamp and command) (listener of ControlHub)
4. Blockly block for setting motor power
5. Blockly block for reading touch sensor (returns fake sensor data)

Changes from last PR:
1. Integrated the ControlHub from a separate repo into this repo.
2. Split UI/View files from ControlHub and moved those to ControlHubView directory
3. Removed redundant state of ControlHub in react comp 'ControlHubView' (which was completely dead and old code)

### DIFF
--- a/src/@types/blockly-javascript.d.ts
+++ b/src/@types/blockly-javascript.d.ts
@@ -1,0 +1,9 @@
+import Blockly from "blockly";
+
+export interface JavaScriptGenerator extends Blockly.Generator {
+  ORDER_ADDITION: number;
+  ORDER_ATOMIC: number;
+  STATEMENT_PREFIX: string;
+  addReservedWords(prefix: string): void;
+  workspaceToCode(workspace: Blockly.WorkspaceSvg): string;
+}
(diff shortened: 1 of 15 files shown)


src/RobotSimulator/StdWorldBuilder.ts
```diff
@@ -0,0 +1,79 @@
+import { Sim3D } from "@fruk/simulator-core";
```
JamesWP, line 1: can you separate this standard world into its own pr?

JamesWP:
This is going to need rebasing on the refactored version of #44 i just finished doing.
It's quite large, it would be better if we can do this in multiple small chunks, each in their own pr. (It's a little hard to review now because of all that, so i'm going to wait till you've had a crack at those)

Issue #25, opened 2020-06-07
Javascript VM should have some consistent execution speed.


JamesWP:
For each given interval of real time, say 1 second. we should execute the running Javascript a given number of steps.

This can be achieved with a set timeout/interval callback to advance the code by a given number of steps.  (would be nice to be able to view the time it takes this to complete).

By advancing the code by a finite number of steps we ensure that any infinite or tight loops the user may come up with dont DOS the rest of the application.

PR #42, opened 2020-06-16
Run VM in a non-blocking way

ginonott:
This PR does some refactoring to run the VM in a way that won't block the main UI thread. I didn't use `requestAnimationFrame` as that runs at a very quick speed and I think theres value in controlling relatively how many blocks we execute per second rather than per frame.

I do not want to merge this without adding test coverage as the logic is starting to get pretty complex between the VM, exposing the VM's state via Redux, and controlling the VM.

It also makes the `vm` the source of truth by holding the state and having the React component simply expose that state globally via redux.


ginonott:
DO NOT MERGE - Waiting on adding test coverage.

ginonott:
Addresses #25

src/JavascriptVM/vm.ts
```diff
@@ -26,20 +48,110 @@ export class BlocklyInterpreter {
...
+      // if paused - schedule the next frame
+      if (this.executionState === ExecutionState.PAUSED) {
+        return this._run();
+      }
+
+      this._step();
+
+      // schedule the next run
+      return this._run();
+    }, 1000 / BLOCKS_PER_SECOND);
```
ginonott, line 116: I think it'd be cool if we also exposed this in some way to the user so they could set the execution "speed" of their program which is why I went with setTimeout and not `requestAnimationFrame` or `setInterval`. Maybe a turtle / hare option? Good follow up issue, nonetheless.

PR #44, opened 2020-06-17
Blockly blocks for Motor and Touch Sensor with dummy VM calls

eurocoinsalbum:


src/BlocklyInterface/scripts/DcMotor.js
```diff
@@ -0,0 +1,57 @@
+import Blockly from "blockly";
+
+Blockly.defineBlocksWithJsonArray([
```
sar-k, line 3: Can this be defined in typescript

src/JavascriptVM/JavascriptVM.tsx
```diff
@@ -88,6 +88,14 @@ export const VMProvider: FunctionComponent = ({ children }) => {
             new BlocklyInterpreter(code, {
               onHighlight: (id) =>
                 dispatch(blocklySlice.actions.highlightBlock({ blockId: id })),
+
+              // dummy implementations
+              onSetDcMotorPower: (port: number, power: number) =>
+                console.log(
```
sar-k, line 94: should this be debug level

src/JavascriptVM/vm.ts
```diff
@@ -23,8 +25,28 @@ export class BlocklyInterpreter {
...
+      const setDcMotorPower = interpreter.createNativeFunction(
+        (port: number, power: number) => {
+          this.pause = true;
+          callbacks.onSetDcMotorPower(port, power);
+        }
+      );
+
+      const isSensorTouchPushed = interpreter.createNativeFunction(
+        (port: number, power: number) => {
+          this.pause = true;
```
sar-k, line 37: Would be helpful to have some comments on the purpose of pause flag
The power parameter is not used. Is it needed?

JamesWP: i think its not required. can likely be removed

src/JavascriptVM/JavascriptVM.tsx
JamesWP: this is a placeholder till we have some way of getting these values to the real simulator

JamesWP: i removed it

src/BlocklyInterface/scripts/DcMotor.js
JamesWP: there is now a type safe way to do this

PR #45, opened 2020-06-17
ControlHub datamodel and listeners

eurocoinsalbum:
independent from any other code


src/ControlHubEmulator/ControlHubEmulator.ts
```diff
@@ -0,0 +1,52 @@
+import { ControlHubConfig } from "./ControlHubConfig";
+import { DcMotorWithEncoder } from "./DcMotorWithEncoder";
+import { ControlHubListener } from "./ControlHubListener";
+
+/**
+ * This class emulates the inputs and outputs of the physical ControlHub
+ */
+export class ControlHubEmulator {
```
ginonott, line 8: I think it would make sense if we use redux for this component. Redux already has mechanisms changing state and maintains listeners which is very similar to what this does. It also has the benefit of allowing other reducers to act on state changes or ignore them.

Thoughts?

PR #47, opened 2020-06-19
Added StdWorldBuilder for the RobotSimulator

eurocoinsalbum:
As requested in an earlier PR I pulled out the StdWorldBuilder


eurocoinsalbum:
Did requested changes. Is it OK that I added vs to gitignore?
StdWorldBuilder (StandardW...) as I expect more predefined WorldBuilders later on before we might switch to load them from a database or URL. It's the one which will be loaded initially when the user opens the website.
The same should be done with the initial robot to allow the user to not only switch the world but also the robot without providing hes/her own one.

.gitignore
```diff
@@ -22,3 +22,4 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 workspace.code-workspace
+/.vs/
```
ginonott, line 25: 👍

ginonott:
Sounds good to me! Maybe the first start of a level builder 😉

Good to merge whenever.

PR #48, opened 2020-06-19
Separate out JS generator type; Add blockly categories

Casper-Oakley:
Cleaned up the blockly toolbox a little to encapsulate different sections in different folders. This doesn't add any new blocks, and doesn't change any block shapes or descriptions or anything, just moves them to categories (in preparation for more to come).

I've also separated out the JS generator type which had been defined in multiple places into a single type.

### DIFF
--- a/src/BlocklyInterface/robotblocks/actuators/DcMotor.ts
+++ b/src/BlocklyInterface/robotblocks/actuators/DcMotor.ts
@@ -1,4 +1,4 @@
-import { addCustomBlock, JavaScript } from "./AddBlockUtil";
+import { addCustomBlock, JavaScript } from "../AddBlockUtil";
 
 export function addDcMotorBlock() {
   addCustomBlock(
--- a/src/RobotSimulator/robotSimulatorSlice.ts
+++ b/src/RobotSimulator/robotSimulatorSlice.ts
@@ -1,10 +1,28 @@
-import { createSlice } from "@reduxjs/toolkit";
+import { createSlice, PayloadAction } from "@reduxjs/toolkit";
+import { RootState } from "../store";
+
+interface IRobotSimulatorState {
+  motors: {
+    [channel: number]: number;
+  };
+}
 
 /**
  * Reducers for handling the state of the simulator such as whether it is executing or not.
  */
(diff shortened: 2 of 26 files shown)


src/BlocklyInterface/robotblocks/actuators/DcMotor.ts
```diff
@@ -43,19 +43,18 @@ export function addDcMotorBlock() {
...
         JavaScript.ORDER_ATOMIC
       );
 
       // convert direction to power sign +/-
-      var isForward = dropdown_direction === "FORWARD";
-      let sign = isForward ? "1" : "-1";
-      var code = `setDcMotorPower(${number_port}, ${sign} * (${value_power}));\n`;
-      return code;
+      const isForward = dropdownDirection === "FORWARD";
+      const sign = isForward ? "1" : "-1";
```
sar-k, line 56: should sign not be a number instead of string

Casper-Oakley: It is a number, though it's going in a string - the output of the func is a string of a js function:
`"1" : "-1"` -> ```setDcMotorPower(1, 1 * (0.1));```

versus if it were attempting to output a string:
`"\"1\"" : "\"-1\""` -> ```setDcMotorPower(1, "1" * (0.1));```

sar-k:
@Casper-Oakley looks good except the unit tests are failing

Casper-Oakley:
> @Casper-Oakley looks good except the unit tests are failing

yep, you're completely right, my bad! I'll push an update with the tests in a sec :)

src/RobotSimulator/robotSimulatorSlice.ts
```diff
@@ -1,10 +1,28 @@
...
  */
 export const robotSimulatorSlice = createSlice({
-  initialState: {},
+  initialState: {
+    motors: {},
+  } as IRobotSimulatorState,
   name: "simulator",
-  reducers: {},
+  reducers: {
+    setPower(state, action: PayloadAction<{ channel: number; power: number }>) {
```
sar-k: nit: You may want to rename the method as setMotorPower to align with the getter

PR #54, opened 2020-06-20
Fix controls mismatched with vm

ginonott:
VM State wasn't being properly synced due to some async wonk-iness and missing code for onFinish.


JamesWP:
Does this solve the issue with the blockly controls? Where they require two clicks to start the code?

ginonott:
Yes! The execution state wasn't being synced correctly from the VM

PR #55, opened 2020-06-20
Styling Improvements - Nicer Motor Controls, CSS Tweaks

ginonott:
# Changes
## Motor Control Group
Rather than buttons and plain text, this uses sliders to control / display the motor. I also added a "lock" button to lock and unlock the sliders so that when a user moves them, they sync together. This makes driving  forward and backwards a little easier until we have actual controller support.

## Styling
I added the FIRST robotics brand logo following this guide: https://www.firstinspires.org/sites/default/files/uploads/resource_library/brand/first-brand-guidelines-2020.pdf


JamesWP:
Can you explain how this handles Being in a locked state when the interpreter changes just one value?

JamesWP:
This looks great though

ginonott:
> Can you explain how this handles Being in a locked state when the interpreter changes just one value?

Right now the lock is only for manually moving the sliders - the VM can set the motors to be whatever it wishes. If still locked and the user overrides the vm then they'll sync together. 

Thoughts?

PR #57, opened 2020-06-20
show messages to the user on errors and successes

ginonott:
This PR introduces message popups. Not all blockly programs produce valid JS code. In this case, the VM may fail to be initialized if it can't be parsed. This displays an error to the user in such a case and sets up a generic container for all types of errors.

To produce an error, add an actuator with no inputs or other blocks and try to start the program.

image

### DIFF
--- a/src/JavascriptVM/JavascriptVM.tsx
+++ b/src/JavascriptVM/JavascriptVM.tsx
@@ -9,6 +9,8 @@ import { blocklySlice } from "../BlocklyInterface/blocklySlice";
 
 import "./JavascriptVM.css";
 import { robotSimulatorSlice } from "../RobotSimulator/robotSimulatorSlice";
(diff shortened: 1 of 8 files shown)


src/JavascriptVM/JavascriptVM.tsx
```diff
@@ -103,8 +105,8 @@ export const VMProvider: FunctionComponent = ({ children }) => {
 
           syncExecutionState();
 
-          setInterpreter(
-            new BlocklyInterpreter(code, {
+          try {
```
JamesWP, line 108: Maybe separate the callback definitions out of the try block? Makes it a little more clear it’s the code we think might fail (creating the BlocklyInterpreter)

JamesWP:
Also the tests failed

PR #61, opened 2020-06-21
Simulator Log View

eurocoinsalbum:
basic table which shows the command which have been sent to the simulator.
Can be extended later on to have a link to the block which initiated the command.
Didn't manage to remove the whitespace in that table
Added react-table module


ginonott:
I think including a table module might be a little over kill. You can render a table like this in React:
```
const data = [{foo: ''} ... ];

return (
  <table>
    <thead>
      <tr><th>Foo Column Header</th></tr>
    </thead>
    <tbody>
      {data.map(d => <tr><td>{d.foo}</td></tr>)} // maps data array to list of react "tr" components
    </tbody>
  </table>
)
```

src/Views/SimulatorLog/SimulatorLog.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useSelector } from "react-redux";
+import { getMotorPower } from "../../RobotSimulator/robotSimulatorSlice";
+
+/**
+ * Component for logging the commands of the simulator.
+ */
+export const SimulatorLog: FunctionComponent = () => {
+  const [tableData, setTableData] = useState<row[]>([]);
+  const motorPower0 = useSelector(getMotorPower(0));
+  const motorPower1 = useSelector(getMotorPower(1));
```
ginonott, line 14: you are right that this isn't really scalable. Instead of building the table in this component, we can add an event log to the robotSimulatorSlice. Then retrieve the list of events with a selector.

It could look something like this:
```
interface IRobotSimulatorState {
    motors: {
      [channel: number]: number;
    };
    eventLog: Array<{eventType: "setPower", /* whatever other values you want ,*/ eventSource: "blockId" /* undefined if set via UI? */}>;
  }
  
  /**
   * Reducers for handling the state of the simulator such as whether it is executing or not.
   */
  export const robotSimulatorSlice = createSlice({
    initialState: {
      motors: {},
      eventLog: []
    } as IRobotSimulatorState,
    name: "simulator",
    reducers: {
      setPower(state, action: PayloadAction<{ channel: number; power: number, source?: string }>) {
        state.motors[action.payload.channel] = action.payload.power;
        state.eventLog.push({eventType: "setPower", eventSource: action.payload.source})
  
        return state;
      },
    },
  });
```

JamesWP:
Does the ui framework we have in the project have a table component built in?

I think we’re using Microsoft’s fluent ui.

Maybe take a look at this?
https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist

eurocoinsalbum:
I don't think that the RobotSimulatorSlice should keep a history of the commands. It's not his job and why should he? If he cares about other view's use-cases he will become too big.
In the log view I also want to show multiple logs. Like from the past n runs. Don't want to put all this burden on the RobotSimulatorSlice.
The other points mentioned I will refactor my code. Thanks for reviewing so fast!

JamesWP:
Redux is designed to store all state, idea idea being without it state is stored in many different places split up in different areas of the application. It becomes very hard to manage.

So if you think separating the log state into different parts is a good idea I’d suggest creating another slice for the state instead of storing it in the component itself.

ginonott:
That makes sense. Would it make sense to simply create a new redux slice
for a log command? We generally don't want react components to keep state
like event history.

ginonott:
> So if you think separating the log state into different parts is a good idea I’d suggest creating another slice for the state instead of storing it in the component itself.

React components can also become unmounted and destroyed, losing all state and this seems like a persistent set of data so its definitely better to hoist it up into redux - but I definitely agree the robotSimulator might not be the best place to put it. Maybe a "monitorSlice" or something?

Here is an example of a reducer that "listens" to actions that belong to a different slice of state: https://github.com/FRUK-Simulator/Simulator/blob/master/src/BlocklyInterface/blocklySlice.ts#L29

In this case, you could add cases to your reducer for each of the robotSimulator commands

eurocoinsalbum:
ok, I think I got the point regarding moving the state into it's own slice away from the component. I will move it to ExecutionSlice which goes into it's own directory.
I just wonder how this would scale if I have 10000 commands and add another one. Will the entire content of the table be reacreated? The listener of the state change can't know which item has been added/changed and therefor needs to recreate all rows from scratch.

JamesWP:
React only create the elements on screen if they need to be created, if the state has one item appended to it then yes the state will be copied momentarily as the master copy is updated.  But react only creates one new HTML element for the new row.

But maybe there would be a limit where we would need to remove old things but I think it’s a problem for future us to solve :D

ginonott:
>  Will the entire content of the table be reacreated?

React has a way of dealing with this as well to determine what in a list needs to be updated in the actual DOM. See https://reactjs.org/docs/lists-and-keys.html#keys. As long as each element in the list

eurocoinsalbum:
thanks for the explanations.
I update the PR and pulled the state out of the component and put it into it's own slice.
Also removed react-table and replaced with simple HTML table.

src/Views/SimulatorLog/SimulatorLog.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const currentExecution = useSelector(getCurrentExecution());
+
+  useEffect(() => {
+    setTableData(currentExecution);
+  }, [currentExecution]);
+
+  return (
+    <table>
+      <thead
+        style={{
```
ginonott, line 20: Can you move the inline styles to `SimulatorLog.css`? I don't think you need to do "display" here.

src/Views/SimulatorLog/SimulatorLog.tsx
```diff
@@ -0,0 +1,48 @@
...
+        }}
+      >
+        <tr>
+          <th>Timestamp</th>
+          <th>Command</th>
+        </tr>
+      </thead>
+      <tbody
+        style={{
+          display: "block",
```
ginonott, line 32: can you move this to the `SimulatorLog.css` file? Also, I don't think you should be overriding the display of the tbody unless you had a specific reason for it?

eurocoinsalbum:
Implemented all required changes from last comment. Thanks a lot - I learned again much :-)

src/Views/SimulatorLog/simulatorLogSlice.ts
```diff
@@ -0,0 +1,53 @@
...
+      .addCase(vmSlice.actions.setExecutionState, (state, action) => {
+        if (action.payload.executionState === ExecutionState.RUNNING) {
+          // creating a new execution log
+          state.logs = [] as ILogEntry[];
+        }
+        return state;
+      }),
+});
+
+export const getCurrentExecution = () => (state: RootState) => state.logs;
```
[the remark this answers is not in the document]
eurocoinsalbum: renamed to getCurrentExecutionLogs in order to indicate that those are for the current/last execution only. Long term we can have more execution logs like from the past n executions.

src/Views/SimulatorLog/SimulatorLog.tsx
eurocoinsalbum: moved to css but whenever I remove display the UI looks weird

src/Views/SimulatorLog/SimulatorLog.tsx
eurocoinsalbum: moved to css but whenever I remove display the UI looks weird

eurocoinsalbum:
edited some comments just right now and uploaded another change

JamesWP:
Two things, the log view should show the most recently executed action, otherwise its just showing stale data always, either reverse the order or scroll to show the last item when you add another one.
i also just noticed that it mentions 'percent' but the value goes between -1 and 1 not -100 and 100.

eurocoinsalbum:
1. new commands will now be added at the top.
2. I normalized the range to use -100 to 100 in the UI and -1 to 1 internally

PR #62, opened 2020-06-22
Simple demo program

eurocoinsalbum:
OPEN FOR DISCUSSION ;-)
This adds per default a simple demo program to the website.
I know the program is not very sophisticated ;-) but it's a starting point.
With one drag'n-drop it could be deleted from the workspace if you want to create your own one.
To create/modify the demo program just implement it in the workspace and click the "run" button.
Die xml of the workspace will be written to the web-console and could then copy-pasted to the code (BlocklyInterface/StdProgramBuilder.ts).
Open for comments and better ideas ;-)
at least the website does not look so empty when loaded initially :-) which has a nice aha-effect


ginonott:
I'm happy that I won't have to keep making programs after every change to the code!

Since I don't think we really want to be editing the xml, does a class to add the xml to blockly seem a little over kill?

eurocoinsalbum:
From Java I'm used to use classes always ;-)
I removed the class and added a function instead.
I added an Interface for BlocklyProgramDemos which can be used to build menus later on.
Functions for loading a program via URL or other sources could be added whenever needed.

JamesWP:
i'm happy for you to merge this :) whenever you're ready

eurocoinsalbum:
...which doesn't make it better ;-) ha ha ha can't stop making those jokes and will never try to, sorry! ;-)
thanks for reviewing

PR #63, opened 2020-06-25
Add reset button

JamesWP:
This also shows a method for accessing things about the simulator from other places.
here i'm using the JavascriptVM Context to act as a middle man.
Its pretty clean (not many lines of code) but i think its a bit messy from an architectural point of view.

Maybe we should think about hoisting the simulator integration up into that JavascriptVM Context. or making some form of amalgamation of them both.

### DIFF
--- a/src/JavascriptVM/JavascriptVM.tsx
+++ b/src/JavascriptVM/JavascriptVM.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, useState, createContext } from "react";
+import { FunctionComponent, useState, createContext, useRef } from "react";
 import React from "react";
 import { useSelector, useDispatch } from "react-redux";
 import { vmSlice } from "./vmSlice";
@@ -26,6 +26,8 @@ export interface IVirtualMachine {
   stop: () => void;
   start: () => void;
   pause: () => void;
+  linkReset(reset: () => void): void;
+  resetSimulator: () => void;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/JavascriptVM/JavascriptVM.tsx
```diff
@@ -50,6 +52,7 @@ export const VMProvider: FunctionComponent = ({ children }) => {
   );
   const dispatch = useDispatch<AppDispatch>();
   const code = useSelector(getCode);
+  const resetLink = useRef<() => void | undefined>();
```
ginonott, line 55: I think this is a good reason for us to hoist the Sim3D instance up to it's own provider or into one that provides both. There's a lot of Refs and I'm a little afraid they'll make debugging harder in the future when we have more than one view as things will get unmounted and destroyed

JamesWP: are you suggesting we do this now? before landing this pr?

ginonott: Yeah, maybe this is a good time for a refactor. Since the VM and the Simulator will need to talk to each other maybe what we can do is have one SimulatorProvider that holds both the interpreter insurance and the Sim3D instance. Then it'd be trivial to connect the two together and for the rest of the application to control them. 

The question is then do we keep redux or simply make the Simulator provider the source of truth for the application?

JamesWP:
The arena switch PR #72 has the same effect as this PR.

PR #64, opened 2020-06-26
Game controller

eurocoinsalbum:
Buttons left side give 50% power and those from right side 100%. This is not end of the game. I will enhance this thingy later on as needed


eurocoinsalbum:
All these are valid points and I will improve it more and more.
But it would be great to have a basic version merged and add features over time otherwise I guess this is a never ending story ;-)
I will also change the game controller's design to have two additional joysticks. Those could be used to move the robot without the motors and set it to a certain position.
3 tabs etc. can go in a seperate ticket - we should wait for some UX feedback before changing too much
Mouse up/down optimizations etc. all this will come over time.
When the code is "good" and "react like" I'm happy if we can merge this PR. Features can be changed/added later.

ginonott:
I agree with making follow up tasks although I don't think we need to wait on UX input to make some layout improvements 🙂

For the code, rather than having 4 functions that basically do the same thing, you can refactor them into 1 handler:

```
const handleMotorChange = (leftPower: number, rightPower: number) => (event) => {
  Dispatch(setLeftAction(leftPower))
  Dispatch (setRightAction(rightPower))
}
```

Use it as such:
```
OnEvent={handleMotorChange (0, 0.5)}
```

You can use function or arrow syntax but I prefer the arrow syntax as I think it's more readable.

src/Views/GameController/GameController.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { robotSimulatorSlice } from "../../RobotSimulator/robotSimulatorSlice";
+import { useDispatch } from "react-redux";
+
+/**
+ * Component for the game controller
+ */
+export const GameController: FunctionComponent = () => {
+  const dispatch = useDispatch();
+  // TODO instead of buttons use the buttons from the control-cross but this path has to be split first.
+  function onBlueButtonClicked(event: React.MouseEvent) {
```
ginonott, line 11: Consolidate into one handler please.

eurocoinsalbum:
I added a service method "handleMotorChange" to control left and right motor at once.
But I don't like to add business code (like handleMotorChange) inside the SVG tags and rather want to stay with seperating UI callbacks and business code.

src/Views/GameController/GameController.tsx
eurocoinsalbum: I added a service method "handleMotorChange" to control left and right motor at once.
But I don't like to add business code (like handleMotorChange) inside the SVG tags and rather want to stay with seperating UI callbacks and business code.

eurocoinsalbum:
I added also now buttons on the steering cross

PR #67, opened 2020-06-26
Add general purpose blocks from FTC blockly implementation

Casper-Oakley:
After spending some time in the FTC blockly implementation, it's becoming clear to me that their blocks won't be importable directly on their own. There's a significant amount of hard-coded calls to Blockly.FtcJava (their Blockly generator), as well as some transpilation from other sources.

However, the general purpose blocks, those that already exist in Blockly land, can just be imported over as they are, so putting in a little PR to cover that and give us some extra basic non-robot-ey blocks while we have a larger think about what robot specific blocks need to be imported over and how best to do that.

### DIFF
--- a/src/BlocklyInterface/robotblocks/general/GeneralBlocks.js
+++ b/src/BlocklyInterface/robotblocks/general/GeneralBlocks.js
@@ -1,0 +1,374 @@
+/*
+ * Code taken and lightly modified from FTC Blockly application. Modifications
+ * cover global variable usage from other files that are not suitable to be
+ * directly copied across, or are themselves transpiled from unknown sources
+ */
+
+// The below lines are the additions to the FTC original general block
+// definition to cover transpiled code parts
+import Blockly from "blockly";
+const miscIdentifierForJavaScript = "miscAccess";
+const functionColor = 289;
+const commentColor = 200;
+
+function createNonEditableField(label) {
+  var field = new Blockly.FieldTextInput(label);
+  field.CURSOR = "";
+  field.showEditor_ = function (opt_quietInput) {};
+  return field;
+}
+
+// FTC copied code begins here
+
+Blockly.Blocks["comment"] = {
+  init: function () {
+    this.appendDummyInput().appendField(
+      new Blockly.FieldTextInput(""),
+      "COMMENT"
+    );
+    this.setPreviousStatement(true);
+    this.setNextStatement(true);
+    this.setColour(commentColor);
+  },
+};
+
+Blockly.JavaScript["comment"] = function (block) {
+  return "";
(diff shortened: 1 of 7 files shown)


eurocoinsalbum:
I think it makes sense (mid term) to support different sets of blocks so that the user can choose.
One set could be for FTC, one for basic robot programming, one for ...
This would help from an educational point of view as well because the pupils would be overwhelmed by too many blocks.
Later on people like teachers should also be able to create their own subsets of blocks and choose those which fit their lessons.
All this not for now but just wanted to highlight this so that it could be considered in the code from the very beginning if possible.

src/BlocklyInterface/robotblocks/general/GeneralBlocks.js
```diff
@@ -0,0 +1,377 @@
...
+};
+
+Blockly.JavaScript["comment"] = function (block) {
+  return "";
+};
+
+Blockly.JavaScript["comment"] = function (block) {
+  return "// " + block.getFieldValue("COMMENT") + "\n";
+};
+
```
sar-k, line 42: Why are there two implementation of this and many other blocks?

sar-k:
@Casper-Oakley - Thanks so much for taking care of it. 
I saw the use of Blockly.FtcJava in the skystone repo too and wasn't sure how to progress.
However, I'll need some explanation on this PR as well as to why there are multiple implementations 
for most blocks and how it all works.
Will check back with you on Monday

JamesWP:
This is looking great, I haven’t had time to test fully yet but I will try to look tomorrow. 
Do we need to add some functions to the JavaScript vm to support some of these blocks? I.e. Math and JavaUtil etc...

src/BlocklyInterface/robotblocks/general/GeneralBlocks.js
Casper-Oakley: Doh! That's my bad, I left in a few Java implementations re-routed to JS, when they should all be JS. Changing now

Casper-Oakley:
Thanks for the comments, gonna re-push in a sec!

> Do we need to add some functions to the JavaScript vm to support some of these blocks? I.e. Math and JavaUtil etc...

Yeah I think there's a "miscAccess" object that the FTC implementation has imported from another file. I believe atan2 and the rounding functions use it for their calculations, so these will need to be imported in / we're gonna wanna figure out a nice way to have some of these functions importable from FTCs implementation so we don't have to keep rewriting our own implementations of blockly block logic - we'll also need to have a think about this area for the non general purpose blocks (motors, atenuators, sensors + others), though we may want to just continue having those separate for now

PR #69, opened 2020-06-28
Issue#52 minimize toolbox

eurocoinsalbum:
Blockly does not allow manipulation of the toolbox except updating the toolbox' xml definition.
You also cannot inject your own elements inside the blockly workspace so the button to toggle min/max has to stay outside
You can see an example on "open roberta" how the toggle between different toolboxes. There the tab-buttons also stay outside and in fact that's the only other example I found where someone tries to change toolbox categories dynamically.
You also can't have more than one toolbox.


src/BlocklyInterface/BlocklyEditor.tsx
```diff
@@ -130,6 +133,27 @@ export const BlocklyEditor: FunctionComponent = () => {
     }
   }, [highlightedBlock, currentBlockSelection, executing]);
 
+  // show/minimize the toolbox. Blockly does not allow the manipulation of the toolbox
+  // in any way except updating the xml definition of it.
+  useEffect(() => {
+    if (blocklyRef.current) {
+      const workspace = Blockly.getMainWorkspace() as WorkspaceSvg;
+      // for an empty toolbox at least one category is needed
+      const emptyToolboxXml = "<xml><category /></xml>";
```
ginonott, line 142: Can we move this to a constant inside of the toolbox file?

src/BlocklyInterface/BlocklyEditor.tsx
```diff
@@ -138,6 +162,10 @@ export const BlocklyEditor: FunctionComponent = () => {
         executing ? "Your program cannot be changed until you stop it" : ""
       }
     >
+      <img
```
ginonott, line 165: Fluent UI has button components that display icons. Could you check https://developer.microsoft.com/en-us/fluentui#/controls/web/button and https://developer.microsoft.com/en-us/fluentui#/controls/web/icon for an appropriate one?

eurocoinsalbum: Nothing nice found which matches in a way that it's intuitive.
I think we can delay beautifying and choosing correct icons until we have an idea about the UI should look like overall.

ginonott: > Nothing nice found which matches in a way that it's intuitive.

A pretty common pattern is to use a Chevron left / right to collapse to the left, expand to the right.

ginonott: > I think we can delay beautifying and choosing correct icons until we have an idea about the UI should look like overall.

Maybe we can set up a meeting with who ever is interested to make some mock ups for how it should look going forward so we don't have a lot of refactoring later on and have an idea of what we're working on.

eurocoinsalbum: Very good idea. We need a concept of how the UI should look like overall at the end.
If the icon is just the last thing which is open from this PR then this shouldn't be a blocker to go forward with this PR, is it?

ginonott: > If the icon is just the last thing which is open from this PR then this shouldn't be a blocker to go forward with this PR, is it?

Can you make a follow up issue for it?

src/BlocklyInterface/BlocklyEditor.tsx
```diff
@@ -130,6 +133,25 @@ export const BlocklyEditor: FunctionComponent = () => {
     }
   }, [highlightedBlock, currentBlockSelection, executing]);
 
+  // show/minimize the toolbox. Blockly does not allow the manipulation of the toolbox
+  // in any way except updating the xml definition of it.
+  useEffect(() => {
+    if (blocklyRef.current) {
+      const workspace = Blockly.getMainWorkspace() as WorkspaceSvg;
+      const blocklyXml = showToolbox ? getToolbox() : getEmptyToolbox();
```
ginonott, line 141: Would it make sense to actually have the redux slice hold the current toolbox and for this component to simply display whatever that toolbox is?

I could see in the future we'd have multiple toolboxes which seems like a piece of business logic / state. 

I'd still keep the exact same action you made and hold whether it's collapsed or not in redux still.

eurocoinsalbum: blockly allows only ONE toolbox :-( 
To have multiple ones you have to change this one to whatever it should display right now. In fact it's just an illusion to the user that he has multiple toolboxes (or collapsed/expanded)

ginonott:
Ah, that makes sense. Would it make sense to have the "current" toolbox XML
string in the reducer (which might be the empty one) and have the component
update it when it changes?

eurocoinsalbum:
Holding the xml definition in the state is not enough for the toggle-visibilty-feature. The onClick has to invert the visible flag. By knowing the current xml-definition it can't really know if the state is collapsed or expanded.
But additional to the visibility flag I added now also the xml definition to the state to support already "multiple toolboxes" and integrated that with the current implementation.

src/BlocklyInterface/BlocklyEditor.tsx
eurocoinsalbum: I created an issue for an overall UI design

Issue #36, opened 2020-06-12
create code view option


JamesWP:
Make it possible to view the generated Javascript in a read only way.

This may require a separate code generator which generates more readable code.

This is to aid Simulator developers, but maybe in the future be turned into a feature for end users.

PR #70, opened 2020-06-28
Add a view for the Blockly-generated Javascript code

sykes-j:
Issue #36 
### Description
- Add a new view, `SourceView`, to display a read-only view of the Blockly-generated Javascript code, with syntax highlighting
- Replace the `Blockly` panel with a new container component, `Editor`, that displays one of the two editor views
- Add a toggle button to the Editor panel to switch between `Blockly` editor and the new `Source` view

### Questions
**Editor state** - I couldn't decide which redux slice would be best to house the new `activeEditor` flag, so I created a new slice for the Editor panel. Would it make sense to put this somewhere else?

**highlightElement lines** - Blockly inserts these `highlightElement` function calls between each line of code. These are probably useful to see for debugging purposes, but clutter the logic otherwise. We could filter these out. Thoughts?

**Toggle button placement** - I've added a "control region" at the bottom of the `Editor` panel to house the new toggle button. I think it makes sense to keep this toggle near the editor, though having a whole control region around it is a little wasteful. If this is an issue I could move this to the top menu bar, or may try to make it float within the editor panel. That being said, if we have more editor controls in the future, a control region would make sense.


src/Views/SimulatorView.css
```diff
@@ -3,10 +3,10 @@
 
   display: grid;
   grid-template-columns: 1fr 1fr;
-  grid-template-rows: 30vw 1fr;
+  grid-template-rows: 60vh 1fr;
```
sykes-j, line 6: This property defines the height of each row, so I thought it was a little strange we were using the view width unit `vw` to define it. Not a biggie either way, happy to revert this.

JamesWP:
This looks fab!

src/Views/SimulatorView.css
JamesWP: this was to ensure the simulator had a consistent aspect ratio as the page resized.

sykes-j:
> This looks good, though im not 100% sure we need a seperate slice in redux for it, maybe we can combine with an existing one? Just an idea.

Having a slice for a single property does feel a bit wasteful. Happy to move this into the blockly slice, if we think that is the best alternative? 

On the other hand, if more editor settings are likely in the future (e.g. show/filter the `highlightElement` in the generated code, some sort of zoom functionality for the editor, colour scheme selection for the blocks) then there's a case for having a separate slice, as these things aren't all completely relevant to the Blockly editor itself.

PR #73, opened 2020-07-04
Add .gitattributes file to ensure consistent line-endings

sykes-j:
In one PR I noticed that a few files that were already on `master` were showing up as changed but with no visible differences. It was probably the line endings that had changed. Adding a `.gitattributes` file will ensure that all line endings are converted to `LF` when checked-in so PRs don't get cluttered with line ending changes.

I haven't created one of these myself before but I've tested myself quickly and it seems to work. Used [this guide](https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings) with more information [here](https://git-scm.com/docs/gitattributes#Documentation/gitattributes.txt-Settostringvalueauto).


ginonott:
Quick question, do any of the files need reformatting to have consistent line endings?

sykes-j:
> Quick question, do any of the files need reformatting to have consistent line endings?

No it looks like they're all LF: e.g. `find src -type f -exec cat -e {} \; | grep ^M`

Turns out the spurious diff we saw on the mentioned PR might have been a rebasing issue rather than line endings, but either way this can't hurt.

Issue #60, opened 2020-06-20
Integrate Google Login


ginonott:
Does not need a backend. Could be clientside only login to start (get the profile from google but do nothing with it).

PR #74, opened 2020-07-05
Integrate Google sign-in on the client

sykes-j:
#60  Add a `Sign in with Google` button that integrates with Google OAuth2 and retrieves some user profile information (name, email, profile photo URL). A `Persona` widget is then displayed with the user's name/profile picture, which has a drop-down button for signing out.

This doesn't implement any session cookies or persist the auth tokens anywhere (just stores in redux state), so upon refreshing the signed-in user details will be lost.

I've created a Google API Credentials from my personal account for testing and committed the client ID so we can play around with it, but I think we'll want to ask First UK to set up and manage the real one going forward. 

 **Testing** 
Google OAuth2 will complain about an unsupported domain name if you're testing this locally. This is because the domain name in the origin/redirects will be `localhost` which is not a domain that can be added to the allow list in the Google API, so it complains the current domain is not allowed. 

Instead you have to "spoof" a real domain, I've set this up with "yourdomain.com". If you're using Unix, you might need to add `127.0.0.1	localhost yourdomain.com` to `/etc/hosts` to set up this resolution. Then changing `localhost:3000` to `yourdomain.com:3000` in your browser should let you sign in.

I've set this up to work with the main netlify domain (https://fruk-simulator.netlify.app/) so it should work on there, but **it probably won't work on the deploy preview links** unless I add each one individually to the allow list. I've add the preview link for this PR to the list so it should work. [You may need to clear your browser cache](https://stackoverflow.com/questions/46472945/idpiframe-initialization-failed-in-google-sign-in-from-localhost) if you're getting sign-in failures.


JamesWP:
Wow, some serious work went into making this work :) very grateful for this!

public/index.html
```diff
@@ -1,20 +1,21 @@
...
+
+<head>
+  <meta charset="utf-8" />
+  <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+  <meta name="viewport" content="width=device-width, initial-scale=1" />
+  <meta name="theme-color" content="#000000" />
+  <meta name="description" content="First UK - Robot Simulator" />
+  <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
+  <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
+  <script src="https://apis.google.com/js/platform.js" async defer></script>
```
JamesWP, line 12: I would hope this can be bundled with the other javascript for our application using webpack. This way we have a chance to deliver a single well cached payload to the browser. And we have full visibility on the JavaScript delivered.

sykes-j: Good catch, turns out we don't need this script tag anyway. The `react-google-login` package manages all the Google OAuth2 stuff for us.

sykes-j:
Fairly confident this won't break the master netlify deployment as the URL is static. Though for the sign-in to work on the future deploy previews, we'll have to manually add each deploy preview link if we want to test sign in. The Google OAuth console doesn't let you use wildcards in the allowed domain list unfortunately. This shouldn't be too much of an inconvenience I don't think.

PR #75, opened 2020-07-06
Move robot simulator state into VM provider

JamesWP:
This hoists up the Sim3D instance into the VM Provider. These two can then more directly interact with each other.

### DIFF
--- a/src/JavascriptVM/JavascriptVM.tsx
+++ b/src/JavascriptVM/JavascriptVM.tsx
@@ -1,4 +1,11 @@
-import { FunctionComponent, useState, createContext } from "react";
+import {
+  FunctionComponent,
+  useState,
+  createContext,
+  useEffect,
+  useRef,
+  useContext,
+} from "react";
 import React from "react";
 import { useSelector, useDispatch } from "react-redux";
 import { vmSlice } from "./vmSlice";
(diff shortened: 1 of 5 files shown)


src/JavascriptVM/JavascriptVM.tsx
```diff
@@ -152,9 +202,31 @@ export const VMProvider: FunctionComponent = ({ children }) => {
             );
           }
         },
+        onCanvasCreated(canvasEl: HTMLCanvasElement) {
```
stevenvaneyck, line 205: Suggestion: Name functions based on what they do, instead of based on what they are triggered on. So in this case, maybe `init`, `initSimulator`?

src/JavascriptVM/JavascriptVM.tsx
```diff
@@ -152,9 +202,31 @@ export const VMProvider: FunctionComponent = ({ children }) => {
...
+          canvasRef.current = canvasEl;
+          updateCanvasSize();
+          // create the simulator
+          sim.current = new Sim3D(canvasEl);
+          const robot = new StdWorldBuilder(sim.current).build();
+          sim.current.beginRendering();
+          robotRef.current = robot!;
+        },
+        onCanvasDestroyed(canvasEl: HTMLCanvasElement) {
```
stevenvaneyck, line 214: Same here. Maybe `destroy`?

src/Views/GameController/GameController.tsx
```diff
@@ -60,6 +62,8 @@ export const GameController: FunctionComponent = () => {
     dispatch(
       robotSimulatorSlice.actions.setPower({ channel: 1, power: leftPower })
     );
+    vm.robot.setMotorPower(0, rightPower);
+    vm.robot.setMotorPower(1, leftPower);
```
stevenvaneyck, line 66: We call this function in multiple places, and each time we also update power on `robotSimulatorSlice`. So maybe we should update `robotSimulatorSlice` from within the `vm` to make sure we always have robot power and the `robotSimulatorSlice` in sync.

PR #80, opened 2020-07-08
Artifact building

JamesWP:


JamesWP:
This job is intended to save an artifact containing the built website, this can then be automatically deployed later.

PR #85, opened 2020-07-09
General Ui improvements

### LINKED ISSUE #76: fix ui sizing issues
after switching the blockly code toggle or hiding the blocks a few times, the page expands and the things at the bottom of the page dissapear.

ginonott:
* moves controller, log, and robot stats to a tabbed view
* created a toggle button bar component
* added hover and active states for the controller buttons for better visual feedback
* added a mouse down state for the dpad to be more inline with a real controller
* fixes infinite page growth bug ( #76 )

FIXES #76


JamesWP:
When I use the preview link, I can see a vertical scroll bar. Something is pushing the bottom of the UI off the visible part of the page. 
I think we should aim to never have vertical scroll bars on the simulator app.

ginonott:
Good catch. I definitely agree although what about stuff like the log?

JamesWP:
As long as they don’t scroll the page itself then we’re good.
The log itself can have its own scroll bar

Issue #78, opened 2020-07-07
add Blockly blocks to query distance sensor values.


JamesWP:

Casper-Oakley:
FTC block for reference (XML can be found in ftc blockly implementation):

![image](https://user-images.githubusercontent.com/5890326/86900951-e3da0400-c103-11ea-8e7e-b4f8701f5762.png)

PR #88, opened 2020-07-21
Add distance sensor

JamesWP:
addresses #78 

This also fixes mixed up left and right controls on the controller.

It also resizes the robot.


JamesWP:
I would be up for extracting the actual implementation from the various fixes and tweaks, but i was too excited to show it working!

JamesWP:
I also noticed the sensor attachment is incorrect ( i think ). @zhiquanyeo  maybe we need an issue for that in SimulatorCore?

JamesWP:
> This looks really good, thanks for the program loaded to give an example. Quick question about the example though: I see the motor movements don't really match the program - is the distance sensor on an unexpected side or something?
> 
> The program ran successfully when I flipped the "Forwards" to "Backwards" and vice versa

Yes there was a bug in Simulator Core which @zhiquanyeo fixed in version 1.3. i've removed my workaround here.

PR #96, opened 2020-07-27
Add challenges with start position and finish zone

stevenvaneyck:
Multiple challenges can be set on the same arena. For now, challenges have the following properties: robot start position and finish zone position. When the robot enters the finish zone, a message is displayed.

Note that arenas and challenges are like DB tables, where challenges have a foreign key into an arena table entry.

Addresses https://github.com/FRUK-Simulator/Simulator/issues/86


stevenvaneyck:
Thx for reviewing! I'm planning to change the size of the arenas to be a bit more realistic and in line with the size of the robot. But I'll do that after this is merged.

PR #99, opened 2020-08-07
Save and load blockly code

### LINKED ISSUE #58: Allow user to save / load a named program
After #32, it would be nice to be able to save multiple programs with a name. Either to local storage or a backend.

MatthiasKreileder:
Closes #58

This PR provides:

* Ability to save the current contents of the blockly editor;
* Load a stored program from the list of available pre-defined and user-defined programs;
* Delete any user-defined programs.

Currently, [localstore](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) is used as the persistence mechanism.  At some point we probably want the blockly programs to be stored on a true backend.


MatthiasKreileder:
@JamesWP @ginonott

JamesWP:
Looks like a great start, Couple of quick things i noticed on first looks:

- The selected arena isn't saved along with the code. (i think maybe this can be done later though)
- The UI needs a little work, it pushes some content off the bottom of the page. (maybe put the save load into a popup?)
- The list of saved programs looks a little funky.

Let me know if you want me to have a look at fixing these parts if you need it :)

MatthiasKreileder:
> Let me know if you want me to have a look at fixing these parts if you need it :)

I am happy to address the points you raised.  I was mainly interested if this goes into the right direction.  I fully agree that the UI needs to made prettier etc.

JamesWP:
would it be possible to also have a download and upload options too. this is something i can look at if you're close to finishing up this pr. or maybe in parallel to this. The ideas would be the user can export the code as well as saving it in local storage. this would then allow them to share the work with the teachers etc...

Please reach out if you you require some help on this issue, it's an important feature!

JamesWP:
Hey Matt just a quick note about the things we spoke about in our catchup:
* The save load ui elements should probably be moved from where they are in the html to be allongside the curent editor and editor controls, this will require using the css grid magic to get the size right.
* The state of the component (the list of saved codes) is currently stored inside the component itself and persisted to local storage, however we discussed moving the state into redux and having redux manage the persistance of the data.

Then I think we call it a good one :) Best of luck

JamesWP (approved):
LGTM, though I removed the lodash throttle stuff, we can add when/if it becomes an issue later

PR #100, opened 2020-08-10
Variable speed

Casper-Oakley:
Thrown together variable speeds in this PR. Contains just a slow and fast toggle.

One issue i'm not sure what the best solution is for is: when the VM executes in a faster state, the end run of a program through the VM + simulator will end in a different location than in the slower state. If for example you have:

```
Motor(100%);
if (true) {
    Motor(0%);
}
```

The presence of the `if(true)` in SLOW will mean that the car in the simulator will be running at 100% speed for an extra like half a second. This means that programs running in fast mode will end at different places to programs running in slow mode. This feels like an issue to me. Anyone have any ideas for what can be done about it?

### DIFF
--- a/src/BlocklyInterface/BlocklyEditor.tsx
+++ b/src/BlocklyInterface/BlocklyEditor.tsx
@@ -82,7 +82,7 @@ export const BlocklyEditor: FunctionComponent = () => {
     if (!blocklyRef.current) {
       blocklyRef.current = new BlocklyInstance(wrapperRef.current!, toolboxXml);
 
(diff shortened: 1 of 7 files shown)


JamesWP:
I think its a good point to make, and it will be very hard to get right too. we would need to sync the speed of the simulator to the code execution speed. For now, I don't think its a huge issue but good to be pointed out.

JamesWP:
Looks like you have a test failure to look into, but other than that looks great.

JamesWP (approved):
Lgtm

Issue #97, opened 2020-07-29
Don't reset arena after program finishes.


JamesWP:

JamesWP:
done!

PR #101, opened 2020-08-14
Halt physics engine instead of re-rendering at program completion

Casper-Oakley:
Rather than re-rendering the vm/simulator, instead request the sim3d instance to disable its physics. This stops movement of any objects, but allows camera movements for further inspection

For https://github.com/FRUK-Simulator/Simulator/issues/97


Casper-Oakley:
Depends on https://github.com/FRUK-Simulator/SimulatorCore/pull/115

JamesWP (approved):
Nice, I believe later we may want to just allow things to continue, but more important for right now
Is just not resetting

JamesWP:
this was implemented in #103 (which contains the commits from this pr)

PR #106, opened 2020-09-11
UI Remaster - Ready For Testing and Review!

ginonott:
# SD -> HD!

This PR will refactor and redesign our UI.

### More self evident directory structure.
* view holds all .tsx (React) files
* state holds all redux code
* core holds all business logic code

view will use code from state and core. state will use code from core. core will be self-contained (should not contain redux or react).

This should make it easier to decide where new code should go.

### Removing Material UI
In favor of making a UI/UX that is a little more friendly and less enterprise-y, I'll be removing Fluent-UI. They'll be replaced with the components / css shown in the mock ups I've made here: https://codepen.io/ginonott/pen/VwaaaPm, https://codepen.io/ginonott/pen/ZEWWYyx, https://codepen.io/ginonott/pen/vYGLzGB

### General Refactors
I'm going to try and re-use as much code as possible as is but will try and make strategic clean ups. In general, I will try and limit this to moving files or removing unused code.

### TODOs
- [ ] Initialize Robot in Redux on challenge load - may become a later issue
- [x] Display Robot sensors on Robot Tab
- [x] Prevent the Simulation from resetting when switching tabs
- [x] Ensure the code executes while switching tabs


Casper-Oakley (changes requested):
Overall, looking really, really nice! I like the look and feel of the whole site. It feels much more like a finished product. What did you have in mind for the images for the levels? Something like a screenshot of each level?

Still working through the code changes themselves. Will get back to you on that, but leaving the below general comments for now:

Nits (levels screen):
1. Sign in with google hover clashes with background on main screen
2. The lesson buttons seem to have a clickable margin to their right, extending the hitbox past the button
3. On widescreen resolutions theres quite a lot of empty space - I idly wonder whether or not its worth having a second column
4. Conversely, the items go off the screen when zoomed in, as opposed to stacking. Probs worth leaving this point for a future item for a better mobile experience; just leaving here for completeness
5. Does FRUK want like a copyright footer or any other footer (or for that matter a link to their site)?

Nits (simulator screen):
1. I'm unable to click the "Actuators" blockly section - looks like the `toolbar-container` is in the way
2. Not really a look/feel issue, but hopping to "My programs" after loading a program, then hopping back to code clears the code
3. I wonder whether or not it'd be worth adding a little tab to the bottom of the toolbar to indicate that there are options there - otherwise its a little unclear until you hover over it
4. In the current master there's a way to delete programs, which this'll need too
5. I wonder if clicking the challenge name at the top should present a dropdown of other challenges. If that's too much/too jazzy then i'd defo make the text un-selectable

ginonott:
> Sign in with google hover clashes with background on main screen

I plan on not touching this very much due to the fact that we really don't use it for anything right now. I think we can make signing in much nicer but we should wait until we have a backend to work. I will fix the text color, though.

> Something like a screenshot of each level?

That was my initial thought but we may need to experiment with it a little bit. Maybe make the levels a little prettier first?

> On widescreen resolutions theres quite a lot of empty space - I idly wonder whether or not its worth having a second column

Thats a good point. Maybe we can try making the section title 100% width, no margins and then the contents can have a reasonable margin on widescreen displays. So it looks more like:
```
--------------------------------
                  Lesson 1
--------------------------------
                   C1 C2 C3
                   C4 C5 C6
--------------------------------
                  Lesson 2
--------------------------------
```

Will try and play around with it. Let me know if you have any other ideas.

4. Agreed.

SimulatorView:

1. CSS is the devil - will try and fix the mouse events.
2. Good catch! 
3. What if we have it start open and then close after a couple seconds then defaulting to its current behavior? 
4. Will add that. I also need to do execution speed and somehow nicely display the robot's motors and sensors. (Ideas very much welcomed!)
5. I'm going to lean against making it a dropdown as I want to be a little more conservative with how many interactive elements appear on the screen at once. Not really sure if its a good idea to disable selecting the text - usually messing with the default browser behavior is anti-user. Not really sure why anyone would wanna copy n paste but 🤷 .

That being said, I was thinking of adding a "Back to lessons" button / shortcut somewhere. Thoughts on where might be a good place for it?

ginonott:
Question: Do y'all prefer the "Show Gamepad" button on the Code screen _OR_ the dropdown in the Settings view?

Casper-Oakley:
Diving in again. To answer your above questions:
 >Do y'all prefer the "Show Gamepad" button on the Code screen OR the dropdown in the Settings view (or something else?)

 I like it how you have it now
> What if we have it start open and then close after a couple seconds then defaulting to its current behavior?

 on initial play, this looks really nice! Defo no need for a tab with that in there
> I was thinking of adding a "Back to lessons" button / shortcut somewhere. Thoughts on where might be a good place for it?

I had actually assumed the first logo was that, as it currently links back to the lessons. Outside of that, can't think of any good spare space for the option, so might be worth just double downing - maybe adding a tooltip to the top left corner logo (unless you plan to make that link to FRUK website or somet)?
> Maybe make the levels a little prettier first?

Yeah good point. Probs worth waiting for James' work to land to improve the visuals around it

Casper-Oakley (changes requested):
Wow what an amazing series of changes, its now looking pretty much feature complete, nice! I've left a few more comments for you to mull over below:

1. Controller is v nice. Only issue I have is if you hold the drag, then let go after a few seconds, it does this weird apparition animation
2. View source was crashing for me. After re-reading these comments its not any more. Not sure what changed but will get back to you if i'm able to reproduce
3. The notification on success/failure's position/size looks a little cluttered. I think its because of the drop-down menu underneath. I wonder if its worth moving it a little lower, or making it a little larger or smaller. In addition, the `X` to close it has an odd colour for the success notification
4. I like the robot status page a lot. Could we make it percentiles rather than decimals?
5. I'm still able to click the "delete" option for the predefined demo programs, even though its greyed out
6. Heads up more than anything, but I see you've got the download option in programs (great!) - we'll also need an "import from file" option. Apols if I've missed it
7. Is it at all possible to make the divider between the simulator and the blockly app draggable at all? Trying on lower resolution and the simulator ends up taking over the screen, making blockly hard to work with otherwise
8. In the lessons screen, the items seem to have a weird layout at my default resolution (fine when zoomed in a little though). Probs want them spaced out more when zoomed out some. See pic below:
image

ginonott:
> Controller is v nice. Only issue I have is if you hold the drag, then let go after a few seconds, it does this weird apparition animation

Thanks! In regards to the apparition, I cant reproduce this on either my windows or mac. It really is ghostly.

> View source was crashing for me. After re-reading these comments its not any more. Not sure what changed but will get back to you if i'm able to reproduce

👍 

> The notification on success/failure's position/size looks a little cluttered. I think its because of the drop-down menu underneath. I wonder if its worth moving it a little lower, or making it a little larger or smaller. In addition, the X to close it has an odd colour for the success notification

I'll keep playing with this. 

> I like the robot status page a lot. Could we make it percentiles rather than decimals?

Good idea!

> I'm still able to click the "delete" option for the predefined demo programs, even though its greyed out

whoops :)

> Heads up more than anything, but I see you've got the download option in programs (great!) - we'll also need an "import from file" option. Apols if I've missed it

I'll add a placeholder for it somewhere.

> Is it at all possible to make the divider between the simulator and the blockly app draggable at all? Trying on lower resolution and the simulator ends up taking over the screen, making blockly hard to work with otherwise

That is an interesting feature request. I'm going to avoid implementing that (for now) but I can create an issue for that. We can either be fancy and allow the user to drag the divide between them or be lazy and making it a setting with predefined size options in the "Settings" tab.

> In the lessons screen, the items seem to have a weird layout at my default resolution (fine when zoomed in a little though). Probs want them spaced out more when zoomed out some. 

I'll keep playing with this too. Might have to cap it off somewhere though and iterate.

JamesWP (changes requested):
I think we might want to change the icons and the text on the simulator controls before we land this.

Start ( > ) < - > Stop ( [.] )
Play ( > ) < - > Pause ( || ) 

Can we also persist the turtle or rabbit setting?

JamesWP (approved):
Let’s do those changes in another pr

PR #107, opened 2020-09-11
Save to file

Casper-Oakley:
Add exporting to JSON + importing from JSON. The files maintain their title, even if renamed, as its stored in a JSON. Currently includes a semver version, coded to 1.0 at the minute. This is also the case for local storage items.

Also added a quick bit to update the current program selection in the dropbox based on what is currently loaded.


MatthiasKreileder:
That's a nice feature!  I played around with the new UI controls for a bit and seems to work just fine!  I did notice one thing that we might want to address though.

If the user wants to export the current blockly program, but didn't supply a program title before clicking on 'Export to File', shall we maybe use the file name as the program title?  Currently the "title" field in the JSON file will be left empty.

The reason I am asking this is because otherwise the program dropdown list looks confusing when a user imports the program that was exported earlier.

Exact steps to trigger this behaviour

1. Start simulator
2. Do not provide a name in 'Program name' input field.
3. Select 'Export to File' => File Dialog comes up
3. Provide a file name, say foobar.json
4. Click 'Save' => File Dialog closes
5. Re-start simulator
6. Click 'Import from File ...' and import foobar.json
7. Look at program dropdown lists, the program doesn't have a title.

Let me know what you think, maybe it's not worth handling this better ...

Casper-Oakley:
Made the above change, lemme know if you find any other weirdness!

JamesWP (changes requested):
Looks good, can we change it so if there is no name we don'd download `json.json` pretty confusing. also we might want to reword the export to file as "Download" and the confusing save locally to "Save"

JamesWP (approved):
LGTM

ginonott (approved):
Nice little change. LGTM.

PR #108, opened 2020-09-13
Scale arenas and add Lesson 3

stevenvaneyck:


src/RobotSimulator/Areanas/lesson3.ts
```diff
@@ -0,0 +1,151 @@
...
+        }
+      } else if (e.zoneId.startsWith("bad-")) {
+        this.actions?.displayMessage("Robot Looses!");
+      }
+    }
+  }
+
+  markComplete(): void {
+    console.log("timeout now");
+    this.actions?.displayMessage("Robot Wins!");
```
stevenvaneyck, line 149: This for some reason doesn't work. The timeout fires (I see "timeout now" on the console) but the message doesn't display.

JamesWP:
Good stuff @stevenvaneyck just need to rebase on master since the new UI came in.

Casper-Oakley (commented):
Scaling looks great! I wasn't able to choose lesson 3 and run my program. When I hit run, it went back to lesson 1. Any ideas why?

Also, i'd reach out to Gino about fixing the merge conflicts. His UI rework has landed, so some stuff has moved around. Shouldn't be an issue looking at your PR, but if you walk into any merge conflicts which you don't expect, i'd ping him a message

src/RobotSimulator/Areanas/lesson3.ts
Casper-Oakley: sounds like this.actions has been reset - possibly the VM has been stopped since, making it undefined?

ginonott:
Merged in changes - just gotta test it.

JamesWP (approved):
LGTM, we need to fix the floor texture before we will properly see the resize :D

PR #111, opened 2020-09-22
Add color sensor

### LINKED ISSUE #79: add color sensor blockly blocks
This would need to return the color of the current simulator core zone the robot is inside.

katgwald:
Fixes #79

Note that this requires the a fix in SimulatorCore to build which is PR'd in https://github.com/FRUK-Simulator/SimulatorCore/pull/124

It also requires a further fix (as yet undiscovered) for the color sensor in SimulatorCore to actually return a value.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1261,13 +1261,20 @@
       }
     },
     "@fruk/simulator-core": {
-      "version": "1.4.1",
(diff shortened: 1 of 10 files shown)


Casper-Oakley (commented):
Code looks good. I'm having trouble getting it to actually work - when I get the robot to go over a green part, and print the colour returned from the sensor, it just comes back with "{}" every time. Is there something i'm missing?

JamesWP (approved):
LGTM

PR #112, opened 2020-09-25
Filter the javascript generated

### LINKED ISSUE #110: filter the javascript generated
filter out all the 'highlightBlock' functions from the code.

MatthiasKreileder:
Filter out all the 'highlightBlock' functions from the generated code.  This makes it easier to read the generated source code.

Closes #110


src/Editor/SourceView.tsx
```diff
@@ -4,6 +4,37 @@ import { getCode } from "../JavascriptVM/vmSlice";
...
+ * @returns A string that is identical to the provided 'blocklyCode',
+ * but without BLOCKLY_HIGHLIGHT_PREFIX statements.
+ */
+function removeHighlightBlockCalls(blocklyCode: string | null): string {
+  if (!blocklyCode) return "";
+
+  const blocklyCodeLines: Array<string> = blocklyCode.split("\n");
+  let output: string = "";
+  for (let i = 0; i < blocklyCodeLines.length; ++i) {
+    if (blocklyCodeLines[i].includes(BLOCKLY_HIGHLIGHT_PREFIX)) {
```
MatthiasKreileder, line 25: This is a simplistic implementation that relies on the `highlightBlock` statements to be on their own line.

I think this comes from two things:
* We append a newline as part of our `STATEMENT_PREFIX`, [here](https://github.com/FRUK-Simulator/Simulator/blob/master/src/BlocklyInterface/BlocklyInstance.ts#L38)
* The blockly generator separates each section via newlines, [here](https://github.com/google/blockly/blob/93b5be40b6229cf164e83b63fe5a6cc4d44a9b5f/core/generator.js#L80-L122)

If we want to want to further future-proof the filtering code we could try to just match  `highlightBlock('*');` or `highlightBlock("*");` via a regex.  It seems the characters that can occur in the unique ID are documented, [here](https://github.com/google/blockly/blob/master/core/utils.js#L423)

Let me know what you'd prefer.

src/Editor/SourceView.tsx
```diff
@@ -4,6 +4,37 @@ import { getCode } from "../JavascriptVM/vmSlice";
...
+ *
+ * @returns A string that is identical to the provided 'blocklyCode',
+ * but without BLOCKLY_HIGHLIGHT_PREFIX statements.
+ */
+function removeHighlightBlockCalls(blocklyCode: string | null): string {
+  if (!blocklyCode) return "";
+
+  const blocklyCodeLines: Array<string> = blocklyCode.split("\n");
+  let output: string = "";
+  for (let i = 0; i < blocklyCodeLines.length; ++i) {
```
Casper-Oakley, line 24: I'd use an iteration function here - in this case `.reduce(` - rather than using simple iteration:

```
const lines = blocklyCode.split("\n");

return lines.reduce((acc, line, idx) => {
    if (line.includes(BLOCKLY_HIGHLIGHT_PREFIX)) {
        return acc;
    }

    acc += line;
    if (idx+1 < lines.length) {
        acc += "\n";
    }

    return acc;
}, "");
```

...or something

Casper-Oakley (changes requested):
I've left a couple of comments below, but I have a more general one:

I hesitate about us doing this kind of simplistic filtering to remove these lines. For starters, I could envisage needing more complex requirements in the future for filtering lines (e.g. what if we wanted to add some other more complex, multi-line internal only formats for our JS output - say for example we needed to add timers to different sections which we didn't want exposed in the output JS). In addition, just running a `.includes(` has its own pitfalls. If I had a function with "highlightBlock" in the name, or a comment or similar, that'd be dropped.

The way to avoid these is to use babel (or a similar tech, I've only used babel before) to parse the output to an AST (sounds expensive, but actually our source is so small that babel will be able to chunk through it in no time) and write a simple babel plugin to remove the lines. With the code in AST, we're able to apply these kind of transformations much more securely, and we don't have to worry about edge cases that simple parsing would hit. In addition, if we needed to apply any other "code output" stuff in the future, we'd have the infrastructure in place to do so.

What are your thoughts? Also worth mentioning that going the babel route could totally not be worth the effort, but if we ever wanna do other forms of output cleanup in future, regexes/string checks potentially won't be sufficient

JamesWP:
Sounds plausible but it’s a problem we can look into when it happens later.

MatthiasKreileder:
Thanks for your detailed comment 👍 

> What are your thoughts? Also worth mentioning that going the babel route could totally not be worth the effort, but if we ever wanna do other forms of output cleanup in future, regexes/string checks potentially won't be sufficient

I am totally up for doing something clever by basing our filtering off the AST structure.  As outlined by you, that is way more powerful than text manipulation.  Nevertheless, to get the conversation going I think it was good to start easy.  Otherwise someone might have said we are over-engineering things for the purpose of fulfilling potential use cases.

I'll close this PR and open a new draft PR once I get something basic working that uses the mechanism you described

JamesWP:
Nooooo Matt! Please land this version first :D

MatthiasKreileder:
Okay, leaving this open for now.  I am fine either way (text manipulation or AST).  Let's see how the discussion evolves.

Casper-Oakley:
Fine with me w/out the babel side of things (and in future we know where we'll need to go if these requirements change!) - could you still make the reduce change Matt?

src/Editor/SourceView.tsx
JamesWP: I think we can happily rely on them being on their own lines :). i think a regex which matches any line where the highlightBlock string appears can be removed. regex: `^\w*highlightBlock.*$` or something like that will work.

JamesWP (approved):
LGTM

JamesWP:
turns out i found a nice way to simplify the code even more:
```JavaScript
  if (!blocklyCode) {
    return "";
  }
  return blocklyCode
    .split("\n")
    .filter((line) => !line.includes(BLOCKLY_HIGHLIGHT_PREFIX))
    .join("\n");
```

PR #120, opened 2020-10-02
fix wheel thickness

JamesWP:


Casper-Oakley (approved):
Ohhh, lovely, looks perfect!

ginonott (approved):

Issue #109, opened 2020-09-18
Add Challenge long description


JamesWP:
maybe with markdown and have this displayed on the page somewhere the students can reference this

PR #123, opened 2020-10-02
Challenge descriptions

Casper-Oakley:
For https://github.com/FRUK-Simulator/Simulator/issues/109

Changed the "return to challenge" option to a new tab. Contains a markdown parsing html blob with a return button inside. I'm not that happy with the look, so suggestions are much appreciated!

Also slightly tweaked the landing page sizes a little. Same as above, any suggestions on the look are welcome


ginonott:
Couple suggestions for you:
* Change the return button to say "<- All Challenges" and I'd suggest adding a color to the back arrow so it pops a bit more
* Move the button from the header to the bottom of the panel and make either centered or full width
* Remove the bottom border of the markdown and make the markdown take all of the available width
* add 1em of padding to the container so that there is some spacing
* Style the code / pre portion of the markdown to have a background and make sure it uses the Roboto Monospace font
* Make sure the rest of the text is using the Roboto font face

Casper-Oakley:
Thanks for taking a look! I've made some of the changes. Got a question about one or two of them:
`I'd suggest adding a color to the back arrow so it pops a bit more` - on just the arrow itself? Any colour in mind?
`make the markdown take all of the available width` - do you mean the bit where the scroll is, or something else?

JamesWP (approved):
LGTM

PR #124, opened 2020-10-02
Keep blockly state when switching tabs

### LINKED ISSUE #115: dont throw away blockly program when switching tabs

MatthiasKreileder:
Store the `BlocklyInstance` state in redux.  Otherwise any unsaved blockly workspace changes are lost whenever the `BlocklyEditor` tab gets unmounted (e.g. when the user switches to a different tab).

Fixes  #115

### DIFF
--- a/src/view/views/MyProgramsView.tsx
+++ b/src/view/views/MyProgramsView.tsx
@@ -35,9 +35,23 @@ export const MyProgramsView = () => {
       dispatch(
         blocklySlice.actions.setActiveBlocklyProgramId({ title: program.title })
       );
+
+      // We need to load the blockly XML that is associated with the
+      // selected 'program.title'.
+      for (const entry of programs) {
+        if (entry.title === program.title) {
+          dispatch(
+            blocklySlice.actions.setBlocklyXmlWorkspace({
+              blocklyXmlWorkspace: entry.xml,
+            })
+          );
+          break;
+        }
+      }
+
       history.replace("?view=code");
     },
-    [dispatch, history]
+    [dispatch, history, programs]
   );
 
   const newProgramCallback = useProgramDialog("create");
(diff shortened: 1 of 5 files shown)


Casper-Oakley (approved):
Looks good and worked great, lgtm

src/view/views/MyProgramsView.tsx
```diff
@@ -35,9 +35,20 @@ export const MyProgramsView = () => {
       dispatch(
         blocklySlice.actions.setActiveBlocklyProgramId({ title: program.title })
       );
+
+      for (const entry of programs) {
```
JamesWP, line 39: i guess the intention is to search for the matching program. but there could be duplicates. if this really is a list then we likely want to match the first instance. if its actually an object and each program a property then we can just programs[entry.title] or similar (i havent looked)

JamesWP:
maybe we dont need the warning about loosing unsaved work?

MatthiasKreileder:
> maybe we dont need the warning about loosing unsaved work?

Thanks for pointing this out.  Yes, I think we can remove the prompt now.  On the grounds that leaving the `SimulatorView` no longer means that we loose the BlocklyInstance state.  As far as I can see there is no other state that is lost when the user leaves the `SimulatorView`.

src/view/views/MyProgramsView.tsx
MatthiasKreileder: > i guess the intention is to search for the matching program.

Yes, the intention is to search for the matching program.   We now match the first instance, as suggested.

> but there could be duplicates

I believe there should be no duplicates.  Because the program title is used as the identifier for the program, so each program title has to be unique.  I think our codebase ensure that this invariant hold.  E.g. if you save the current code and give it a title that is already taken then you'll save over the existing program, as opposed to creating a second program with the same title.

I guess the underlying issue here is that we store the programs in an array, instead of a map from program id (title) to program objects (description and xml-encoded blockly code).  That way we could just do map look-ups instead of looping through our array.

PR #125, opened 2020-10-16
Use the custom robot from SimulatorCore

Casper-Oakley:
Just put in the import line for the robot. Included the 120KB robot sent around via email, looks good to me as is.

I guess there's still an open question about grabbers. https://github.com/FRUK-Simulator/SimulatorCore/pull/127/files has the grabber as a bucket facing in front of the robot, leaving the arm at the top of the FRUK robot looking a little out of place

Source: https://github.com/actions/setup-node/issues/212

PR #126, opened 2020-11-13
Add grabbing block

Casper-Oakley:
### Description

Add a block for sending grab opens/closes to the robot. Based off of the WIP https://github.com/FRUK-Simulator/SimulatorCore/pull/127 - this PR is subject to change from that. Have only put in one for inputs, and have left out checking for now. Will put that in with this PR shortly

### Dependencies

https://github.com/FRUK-Simulator/Simulator/pull/125 - Robot model import
https://github.com/FRUK-Simulator/SimulatorCore/pull/127 - Grabber functionality in simulator


src/RobotSimulator/StdWorldBuilder.ts
```diff
@@ -74,7 +74,44 @@ export class StdWorldBuilder {
...
+        depth: 1,
+        maxWidth: 2,
+        minWidth: 0.9,
+        mountOffset: { x: 0, y: 0, z: -0.01 },
+        mountFace: RobotSpecs.SensorMountingFace.FRONT,
+      },
+    ];
+
     robotSpec.initialPosition = this.startPosition;
+    robotSpec.customMesh = {
```
JamesWP, line 101: looks like this contains the changes from another pr

src/BlocklyInterface/robotblocks/mechanisms/MechanismInput.ts
```diff
@@ -0,0 +1,33 @@
+import { addCustomBlock } from "../AddBlockUtil";
+
+export function addMechanismInputBlock() {
```
JamesWP, line 3: The blocks should more closely map to the set and get digital commands. The fact that the channel is actually controlled by the grabber mechanism is abstracted from the programmer in this case

JamesWP (changes requested):
I believe we want the blockly blocks to be set and get digital values, having the mechanism be abstracted from them

Casper-Oakley:
I had actually originally implemented it that way, but it felt kinda confusing for the end user. Like, if I wanted to program the robot to pick up a block I'd need to know what digital channel the robot grabber is on as well as what channel to read from. These aren't indicated anywhere in the application at present, and are instead just hard-coded. If you wanna make it a direct digital interface, should these also be indicated somewhere? Same applies to motors imo

Issue #122, opened 2020-10-02
move popup messages somewhere they wont get in the way of the tabs


JamesWP:
the robot wins! and robot looses! messages overlay the tabs, they also dont fade after a time so you are forced to close every one of them

PR #127, opened 2020-11-17
Optionally fade out display messages (user notifications)

MatthiasKreileder:
Arenas/lesson1.ts: Make success/failure messages fade away after a configurable amount of time (default 2s)

Extend the `ChallengeActions` interface so that it provides an option to display messages that fade out after a configurable amount of time.

Related to #122


JamesWP (approved):
LGTM

PR #129, opened 2020-11-20
Integrate grabbing (re-built)

Casper-Oakley:
### Description
(Re-make for https://github.com/FRUK-Simulator/Simulator/pull/126)

Add a block for sending grab opens/closes to the robot. Based off of the WIP https://github.com/FRUK-Simulator/SimulatorCore/pull/127 - this PR is subject to change from that. Have only put in one for inputs, and have left out checking for now. Will put that in with this PR shortly

### Dependencies

https://github.com/FRUK-Simulator/SimulatorCore/pull/127 - Grabber functionality in simulator


src/RobotSimulator/StdWorldBuilder.ts
```diff
@@ -74,6 +74,30 @@ export class StdWorldBuilder {
...
 
     const robotSpec = robotBuilder.generateSpec();
+
+    robotSpec.mechanisms = [
+      {
+        type: "gripper-mechanism",
+        ioMap: [
+          {
+            id: "grab",
+            channel: 0,
```
JamesWP, line 84: I think some other sensors might conflict with these values. but i might be wrong there.

Casper-Oakley: Checking now. I had assumed that they were different channels types, but will step through on the simulator core side. Worst comes to worst ill just bump it up to unique numbers

Casper-Oakley: Looks like they are considered separate systems, and don't conflict. Want me to increment them anyway so they're unique for the end user or just leave as is?

JamesWP:
Either or, (ill approve now) but if its not changed then add a comment about that

JamesWP (approved):
LGTM

Casper-Oakley:
Will leave as is and come back to it if there's an issue

PR #132, opened 2020-12-18
Add challenge results display

### LINKED ISSUE #114: Add challenge results display
Next to the challenge we should display the current state. pending, failed success, this should reset each time the arena is reset. and the challenge code itself should be able to mark a challenge pending failed or success (only the first update should be shown and subsequent ones ignored).
when the result failure or success is determined we can switch to showing the page with the info on.

MatthiasKreileder:
Show the current challenge state in the `LandingView.tsx` and the `ChallengeView` component.

To be able to show the challenge state this commit:
* adds an additional slice, `challengeSlice.ts`, that stores the current state of each challenge.  Challenges can be in one of three states: Pending, Success and Failure.
* Extends the `ChallengeActions` interface to provide lessons an API to mark the current VM run as success/failure.

Closes #114

### DIFF
--- a/src/JavascriptVM/JavascriptVM.tsx
+++ b/src/JavascriptVM/JavascriptVM.tsx
@@ -32,6 +32,10 @@ import {
 } from "../RobotSimulator/Arenas/base";
 import { ChallengeActionsImpl } from "./ChallengeActionsImpl";
 import { getDefaultChallenge } from "../RobotSimulator/ChallengeConfigLoader";
+import {
+  challengeSlice,
+  ChallengeStatus,
+} from "../RobotSimulator/Arenas/challengeSlice";
 
 /**
  * Interface to control the VM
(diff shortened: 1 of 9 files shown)


src/RobotSimulator/Arenas/challengeSlice.ts
```diff
@@ -0,0 +1,57 @@
...
+  id: string;
+};
+
+export const challengeSlice = createSlice({
+  initialState: {
+    challengeInfos: [] as ChallengeInfo[],
+  },
+  name: "challenge",
+  reducers: {
+    setChallengeStatus(
```
MatthiasKreileder, line 21: The update logic in this reducer is probably a bit hard to read.  I need to think further how it could be simplified.  Maybe we could pass the current challenge state into the `ChallengeActionsImpl` constructor, and then the ChallengeActionsImpl object can report success/failure based on the events it receives and the current challenge state.  That way there would be less logic in here ...

src/view/views/ChallengeView.tsx
```diff
@@ -31,6 +55,8 @@ export const ChallengeView = () => {
         classes={["challenge-view--content"]}
       />
       <Divider />
+      <p>Status: {rst}</p>
+      <Divider />
```
MatthiasKreileder, line 59: This should probably be an icon in green/red/grey based on the status or something.

Probably we also want to show this state in the challenge overview page.

src/JavascriptVM/JavascriptVM.tsx
```diff
@@ -310,9 +314,21 @@ export const VMProvider: FunctionComponent = ({ children }) => {
             challengeListener.current.onStop();
           }
 
+          // Before we start the challenge listener we mark the challenge
```
MatthiasKreileder, line 317: One disadvantage of this approach is that it we might "loose" the state of a session when it is currently marked as failure.  E.g. say a lesson has been marked as a failure, and the users starts over and runs the code again, and this time the code doesn't produce anything that triggers a 'failure' event (e.g. the robot did not drive into a bad zone) then the lesson state goes from 'failure' to 'pending'.  Maybe that's okay.

Alternatively, instead of resetting the state here, we could pass the current state  to the ChallengeActionsImpl object, and have the ChallengeActionsImpl figure out the details of how to correctly update the slice.  The ChallengeActionsImpl gets notified of all the events happening during the current run, together with the status of the challenge at the beginning of the run it can then update the slice.

JamesWP (approved):
LGTM

MatthiasKreileder:
Thanks for your review and the discussion, @JamesWP.  I'm happy to iterate on this and do some more follow-up work as we get more feedback and the project evolves.  I'm planning to land this first version now.

PR #140, opened 2021-04-25
add debug mode to show state of physics world

JamesWP:
This is only intended for debugging the simulator.

### DIFF
--- a/src/view/views/SettingsView.tsx
+++ b/src/view/views/SettingsView.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect } from "react";
+import { useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { loadSettings, persistSettings } from "../../core/settings/settings";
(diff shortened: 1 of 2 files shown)


src/view/views/SettingsView.tsx
```diff
@@ -98,6 +112,27 @@ const CameraViewSelect = () => {
   );
 };
 
+const PhysicsDebugSelect = () => {
+  const vm = useVM();
+  const [physicsDebug, setPhysicsDebug] = useState(vm.isDebugMode());
```
MatthiasKreileder, line 117: I suppose you deliberately chose `useState` over `useSelector` so that you don't have to make this part of a slice?   The other settings options are implemented via `useSelector`.

MatthiasKreileder (commented):
Nice that this debug option becomes available in the Simulator.

JamesWP:
Yes it’s laziness on my part, we can make it persistent in later prs I think

MatthiasKreileder (approved):

PR #147, opened 2021-06-18
vm: fix early exit on wait

### LINKED ISSUE #149: Wait command does not work correctly if no instruction after it (HIGH priority)
Wait command doesn’t seem to work unless it has a block after it? E.g. “wait 10” at the end is skipped over, but “wait 10” and then “wait 0” works. It works if a block is put after it and then deleted?

JamesWP:
We should exit from running statements if we are supposed to be waiting.
Without this, a program which ends on a:
wait(n); 
will simply exit before waiting.

Fixes #149

### DIFF
--- a/src/JavascriptVM/vm.ts
+++ b/src/JavascriptVM/vm.ts
@@ -268,7 +268,7 @@ export class BlocklyInterpreter {
   private _step(): boolean {
     let finished = false;
     this.blockHighlighted = false;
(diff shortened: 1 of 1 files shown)


MatthiasKreileder (approved):
Nice, seems to work well.  Also the highlighting looks better now, because the wait block is being highlighted while we are waiting.

Issue #153, opened 2021-06-22
Level should not be passed if black walls hit (LOW priority)


gavcradd:
Level should not be passed if the black boxes are hit on the way there (at the moment it says “failed” but this is then overridden with a pass when the green box is reached). Perhaps use a flag to record whether the black boxes have been touched and only display the success message if the robot makes it to the end and the flag is still set to FALSE/NULL.

PR #162, opened 2021-07-02
Improve success/failure report messages

MatthiasKreileder:
Relates to https://github.com/FRUK-Simulator/Simulator/issues/153

- Don't emit a 'Robot Wins!' message if the robot has crossed a "bad" zone during the current run.
- Similarly, don't emit a 'Robot Looses!' message if the robot has already reached the 'FinishZoneId' cleanly (i.e. without driving into a bad zone beforehand).

If we are okay with the proposed approach then I can apply this to the other lessons.


JamesWP:
The below program reproduces the issue.

[finish after fail.json.txt](https://github.com/FRUK-Simulator/Simulator/files/6765127/finish.after.fail.json.txt)

JamesWP (approved):
LGTM

PR #164, opened 2021-07-11
Show minimal blockly code at the first launch

### LINKED ISSUE #148: Remove default code when simulator is loaded (LOW priority)
At the moment, the simulator loads with the "Crazy Robot" program. Please remove this so that the code window is empty.

MatthiasKreileder:
Create a new, minimalistic, blockly program that only sets the motor power on one side.  Make this new program the default code that gets loaded on the very first start of the simulator. 

Note that the current logic of handling blockly code (persisting code in local storage across re-loads and challenges etc) remains the same.

Relates to #148 - we can also start with a truly empty blockly editor?  Maybe that's better?  I personally think that showing a minimal, runnable, example is better than no blocks at all, because it provides a more intuitive starting point.  Let me know what you think and I'll update this PR accordingly.

### DIFF
--- a/src/core/blockly/programs.ts
+++ b/src/core/blockly/programs.ts
@@ -31,6 +31,13 @@ export const newProgramXML = `<xml xmlns="https://developers.google.com/blockly/
  */
 
 export const predefinedDemos: BlocklyDemoProgram[] = [
+  //
+  {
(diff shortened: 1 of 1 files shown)


JamesWP (approved):
LGTM

PR #169, opened 2021-07-19
Add Block to read gyroscope and show rotation in robotview

MatthiasKreileder:
The `StdWorldBuilder` now adds a gyroscope sensor by default.  Demo program that can be imported: https://gist.github.com/MatthiasKreileder/65809c4cc5f314bb306ad3a10844ef2f

There's one problem I observed: When we execute the blockly code in 'slow' mode then we might "undersample" the sensor.  E.g. when you turn the robot right until the gyro shows 90 degrees then you'll most likely overshoot that value by a few degrees.  That's because in slow execution mode the sensor value doesn't get read very often.  When you run the sample code with the fast execution speed then the movement is very accurate, because the undersampling problem goes away.  I think the same applies to other code that wants to base logic off the sensor reading (like distance sensor readings) - it's nothing specific to the gyroscope in my understanding.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@fluentui/react": "^7.168.2",
-        "@fruk/simulator-core": "^1.6.4",
+        "@fruk/simulator-core": "^1.6.6",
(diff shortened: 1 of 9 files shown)


JamesWP (approved):
Lgtm

PR #171, opened 2021-08-03
Add start block

JamesWP:
This allows only the code inside the start block to be run.

In future we can remove from toolbox and make it unremovable.

It also unifies all the hacks around the typescript access for Blockly.JavaScript


src/JavascriptVM/vm.ts
```diff
@@ -233,7 +233,12 @@ export class BlocklyInterpreter {
       );
     });
 
+    // this may cause an infinite loop, we should bail out if this is the case
+    this.interpreter.run();
+
     // Start running in a paused state - ie, spawn a "thread"
+    this.interpreter.appendCode("start();");
```
MatthiasKreileder, line 240: If someone 'load's code that does not define a `start()` function, then nothing happens. Should we instead maybe emit a warning?  I am aware that implementing such a check in a way that works in all possible scenarios might be tricky (or an unreasonable amount of effort) - but given that we control the code that gets emitted for a start block we might be able to come up with a check that should work in most (almost all?) cases?

src/JavascriptVM/vm.spec.ts
```diff
@@ -11,7 +11,7 @@ describe("javascript vm", () => {
   });
 
   it("calls the highlightBlock callback", () => {
-    const code = "highlightBlock('abc');";
+    const code = "function start() {highlightBlock('abc');}";
```
MatthiasKreileder, line 14: Similar to the required updates to the tests, I think it would be best to update the example programs, so that they can be loaded and run w/o modification? If you prefer that we do this via a follow-up PR then that's fine too.

src/JavascriptVM/vm.ts
```diff
@@ -233,7 +233,12 @@ export class BlocklyInterpreter {
...
+
```
MatthiasKreileder, line 241: I think it would be nice if the 'view source' component also shows that 'start()' gets called - currently
it looks like we just define a function, without running it.

src/JavascriptVM/vm.ts
```diff
@@ -233,7 +233,12 @@ export class BlocklyInterpreter {
       );
     });
 
+    // this may cause an infinite loop, we should bail out if this is the case
```
MatthiasKreileder, line 236: I am afraid this could be rather common, e.g.:

unresponsive

Would it be possible to step() the code until the JS-Interpreter's `step()` method returns false? We could use a counter to "bail out" if we haven't managed to parse the code to completion within a reasonable amount of steps?

MatthiasKreileder (commented):
Thanks!  I left a few comments, let me know what you think.

src/JavascriptVM/vm.ts
JamesWP: I'm not sure how to start this one. I think we can wait till we've seen this happen first.

src/JavascriptVM/vm.ts
JamesWP: i've managed to fix this, now only code in Start will be generated. I still need to handle the case where there is no start. Ive put some tokens in the generated javascript to detect later.

src/JavascriptVM/vm.spec.ts
JamesWP: done!

src/JavascriptVM/vm.ts
JamesWP: that might be ok, think of it like `main();`

PR #195, opened 2021-12-12
Add limit to number of blocks

### LINKED ISSUE #92: Add limitation for number of Blockly blocks
Some challenges require this. For now just on Blockly side.

Later we can think of integration with arena/challenge selection

ido-ran:
Fix #92 by allow challange config to specify the number of blocks allowed
to be used. This should help students to identify the are on the wrong
path but it seems like a clear feedback of how many blocks are allowed
is required as blockly simply turn all blocks into "disabled" without
clear explanation why that is.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@types/marked": "^1.1.0",
         "@types/react-syntax-highlighter": "^11.0.5",
         "@uifabric/icons": "^7.5.23",
-        "blockly": "^3.20200924.4",
+        "blockly": "^6.20210701.0",
         "js-interpreter": "^2.3.0",
         "marked": "^2.0.3",
         "normalize.css": "^8.0.1",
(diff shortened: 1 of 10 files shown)


JamesWP:
Great start! I think there needs to be some on screen display for the limit, the blockly interface isnt very informative about why you cant select any more blocks. maybe some counter on screen when you approach the limit, this way there is some place to put a reason.

ido-ran:
I've converted this PR to draft because we will need to remove the limits as I'm not sure they are correct.

ido-ran:
I have add the ability to specify if the limit is hard or soft and also more descriptive messages about the number of blocks.
Please see lesion 1 challenge A which has max 6 soft limit and challenge B which has 6 hard limit.
Try to use more blocks then expected in the soft limit.

You can still load a program with more than the allowed limit, even when the limit is hard.
I'm not sure if there is a point in blocking that.

JamesWP:
Works great! I think this is in a place we can land it, there's a chance to improve the looks but that can wait IMO. (I think we could wrap that text in one of those box things with the drop shadow. )

Issue #189, opened 2021-10-15
Lesson 2 challenge completion message is missing on completion


JamesWP:
need to reproduce this, it might just be some missing code in the lesson event handler

ido-ran:
It seems the problem is the `setTimeout` callback is using a different `this` point so it cannot access the `actions` property of the instance.

https://github.com/FRUK-Simulator/Simulator/blob/938de0f462acdcac11e3dfa327912ed536d6b34f/src/RobotSimulator/Arenas/lesson2.ts#L205-L229

ido-ran:
PR #197 is the simplest fix for this issue but I think we should not use `clearTimeout` but instead keep some running counter which is updated whenever we go in or out of the zone.

ido-ran:
@JamesWP this should be closed as it was fixed. Thanks.

PR #197, opened 2021-12-19
Show success message to lesson 2

ido-ran:
This is the simplest fix to issue #189.
The problem was that `this` set by `setTimeout` is not the instance of the class that set it.
Using `bind` we change the `this` to the the class instance.


JamesWP (approved):
Thats a subtle catch, nice work

Issue #187, opened 2021-10-15
Lesson 1 Camera position update


JamesWP:
For lesson 1 ( all challenges ) the camera should have a birds "eye view" i.e. the camera should start above the playfield and look directly down on the robot.

The same camera controls we have now should still allow the camera to be positioned with the mouse.

JamesWP:
i'm going to look into this one

PR #210, opened 2022-02-09
Camera angle and fixes

JamesWP:


src/JavascriptVM/JavascriptVM.tsx
```diff
@@ -170,7 +170,59 @@ export const VMProvider: FunctionComponent = ({ children }) => {
     return () => {
       console.log("Destroying VMContext.Provider");
     };
-  });
+  }, [canvasRef]);
+
+  function setSimulatorCameraMode(
+    simulator: Sim3D,
+    robotRef: Handles.RobotHandle,
+    cameraMode: CameraView
```
ido-ran, line 178: There is a mismatch here with cameraMode and CameraView.
I think we should rename CameraView into CameraMode.

ido-ran:
The new top-down values of x:0 y:7 z:3 produce a view which seems a bit far but does present 3D view.
image

What do you think about using x:0 y:4 z:0 for a complete top-down view
image

src/JavascriptVM/JavascriptVM.tsx
```diff
@@ -39,6 +39,7 @@ import {
   challengeSlice,
   ChallengeStatus,
 } from "../RobotSimulator/Arenas/challengeSlice";
+import { SimBasicSensor } from "@fruk/simulator-core/dist/engine/objects/robot/sensors/SimBasicSensor";
```
ido-ran, line 42: Why was that import added?

ido-ran (commented):
This fix is for https://github.com/FRUK-Simulator/Simulator/issues/187, right?

It is more bird-eye but do you want to make it completely top-down?

src/JavascriptVM/JavascriptVM.tsx
JamesWP: The parameters here are slightly confusing! but i wont try to fix that in this PR

src/JavascriptVM/JavascriptVM.tsx
JamesWP: That was a mistake, i think VsCode has a habbit of adding these when they're not needed! :)

PR #211, opened 2022-02-11
Change navigation to stay static

### LINKED ISSUE #207: Change navigation to stay static
Based on user feedback, please can the bar which has "Code, My Programs, Robot, Settings, Challenge" stay static.

anap1003:
Closes #207 

A couple of TODOs:

- [ ] Figure out how to shrink the toolbar with the left side panel
- [ ] Check if more code is unused and can be deleted

### DIFF
--- a/src/view/views/SimulatorView.css
+++ b/src/view/views/SimulatorView.css
@@ -36,6 +36,7 @@
 /* Left Panel */
 .simulator-view--panel:nth-child(1) {
   flex: 2;
+  margin-top: var(--toolbar-height);
 }
 
 /* Right Panel */
--- a/src/variables.css
+++ b/src/variables.css
@@ -57,12 +57,8 @@
 
   /* Toolbar */
   /* The total height of the toolbar */
-  --toolbar-height: 90px;
+  --toolbar-height: 50px;
   --toolbar-width: 50%;
-  /* How much of the toolbar is visible when 'hidden' */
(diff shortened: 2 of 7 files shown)


JamesWP:
I like it!

JamesWP:
Regarding your second todo point, i think the current sizing is actually better than if it was bounded by the sidebar, the only issue is when it runs out of room to shrink it curently slides off the left edge. instead if it starts to overlap the right edge that would be better since you could still see all the buttons. We could also considere an `@media query` to make it look different when shrunk like that.

src/view/views/SimulatorView.css
```diff
@@ -36,9 +36,16 @@
...
   flex: 2;
+  margin-top: var(--toolbar-height);
 }
 
 /* Right Panel */
 .simulator-view--panel:nth-child(2) {
   flex: 2;
 }
+
+@media only screen and (max-width: 1500px) {
```
anap1003, line 47: @JamesWP is there a way to use a `var` here for `max-width` value so that we can keep the 2 `@media` queries in sync?

anap1003: > This looks great, Is there anything you want to do before we land this to master

@JamesWP I would've liked to couple the 2 `@media` values for `max-width` somehow, but I'm not too fussed about it. Good to land on my side.

Issue #215, opened 2022-03-09
making pause button also freeze time in the simulation, and resume later


JamesWP:

Allegro-Leon-Li:
@ShroukAshraf @JamesWP let's catch up on this sometime later. Should this work be based on Shrouk's branch?

ShroukAshraf:
I believe it should be different branch  @Allegro-Leon-Li  what do you think and sorry for the delay

PR #218, opened 2022-03-29
Pause the simulation when 'Pause' button is clicked

Allegro-Leon-Li:
This is part of the https://github.com/FRUK-Simulator/Simulator/issues/215
The physical world is not paused when pause button is hit. This PR is to make the arena freeze when paused. When the timer feature is done we can make sure both timer and the arena are paused.

### DIFF
--- a/src/JavascriptVM/vm.ts
+++ b/src/JavascriptVM/vm.ts
@@ -37,6 +37,16 @@ export type BlocklyInterpreterCallbacks = {
   onControllerKeyCheck?: (key: ControllerKey) => boolean;
 
   /**
+   * Called when the simulation is supposed to be paused therefore stop all physics
+   */
+  onPause?: () => void;
+
+  /**
+   * Called when the simulation is supposed to resume from pause therefore start all physics
+   */
+  onUnPause?: () => void;
+
+  /**
    * Gets the value of the given sensor on the curent robot. value is between 0.0 and 1.0.
    */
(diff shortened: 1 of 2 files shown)


Allegro-Leon-Li:
I think this PR might be a bit overlapped with the timer wip feature, in terms of the pause button. I created the callback functions for that but let me know if we have other opinions on how this should be implemented.

src/JavascriptVM/vm.ts
```diff
@@ -278,6 +288,7 @@ export class BlocklyInterpreter {
   private _step(): boolean {
     let finished = false;
     this.blockHighlighted = false;
+    this.callbacks.onUnPause && this.callbacks.onUnPause();
```
ShroukAshraf, line 291: I have a question why do you think this line is needed? Don't you have to call the function `unpause` in order for `_step` to work again or I am missing something?

ShroukAshraf (commented):
Thanks @Allegro-Leon-Li for the PR, just asked one question about it.

src/JavascriptVM/vm.ts
Allegro-Leon-Li: If we call `unpause()` it will change the state from `PAUSED` to `RUNNING`, which is a change of current behavior and I believe we don't want that state change.

JamesWP (approved):
LGTM

Issue #193, opened 2021-10-19
Lesson 4 B, C 


JamesWP:
Black portions are no-go zones, touching them ends the challenge.

Challenge B - includes walls, contact should also end the challenge
![image](https://user-images.githubusercontent.com/707236/137945622-db53b748-fc26-4be5-90f0-c8770875c188.png)

Challenge C
![image](https://user-images.githubusercontent.com/707236/137945490-262b9367-8b68-4ab4-8e99-e7afe1114dfe.png)

ido-ran:
Challange B will be mostly about distance sensor, maybe the course should be a bit more similar to A with walls?

JamesWP:
Challenge B will now be modified to not contain the walls shows in the image and instead require the use of the color sensor to detect the green areas and turn 90 degrees. in addition to this we should consider disabling the use of the distance sensor in this lesson completely (maybe a popup if they use it) also consider adding a coloured  block for the end zone which they must also detect.

PR #221, opened 2022-04-10
Add Lesson 4 #193

ido-ran:
Add 2 challenges for lesson 4 to use the color sensor.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,6 @@
         "react-router-dom": "^5.2.0",
         "react-syntax-highlighter": "^15.4.3",
         "redux": "^4.1.0",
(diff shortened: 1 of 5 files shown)


ido-ran:
I've noticed something seems strange with the color sensor events from plank.
When I've activated the Tick Physics debug option I've noticed that the definition of polygon zone for threejs seems to be very different from that of plank.
The following image shows it.

image

As you can see from the image, while the threejs has a "step" at the top left corner, plank has a direct line connecting the two corners.

JamesWP:
Maybe we should add some algorithm to turn these non convex polygons into a series of smaller convex polygons, there are a few algorithms which can do this iirc

PR #249, opened 2022-09-29
Add the ability to configure whether robot has a distance sensor

### LINKED ISSUE #244: Expose robot config as part of arena config
Some of lessons 4 need to have the distance sensor disabled, so we should expose robot configuration as part of arena configuration ideally.

anap1003:
Lessons 4 need to have the distance sensor disabled, so we're exposing robot configuration as part of challenge configuration. 

Let me know if we want to tie the robot configuration in with the arena configuration as originally intended, it made more sense to me to make it a separate config.

Closes #244


JamesWP (approved):
LGTM

PR #259, opened 2022-11-21
Show motors from robot

ido-ran:
Before this change the information about the motors in RobotView was captured when setting the motors and not from the robot itself.

This now uses the RobotSpec and SimRobot, like the rest of the sensors, to decide what to present and get the actual values of the motors.

https://user-images.githubusercontent.com/187938/205295099-7e4cf85d-6f59-4668-8a54-9b53ebbd34e9.mp4

Depends on https://github.com/FRUK-Simulator/SimulatorCore/pull/194

### DIFF
--- a/src/JavascriptVM/JavascriptVM.tsx
+++ b/src/JavascriptVM/JavascriptVM.tsx
@@ -101,26 +101,6 @@ export const VMProvider: FunctionComponent = ({ children }) => {
 
   const store = useStore();
 
-  // handler for robot handlers, all calls to setMotorPower will update state in redux
-  const robot_handler: ProxyHandler<Handles.RobotHandle> = {
-    get: function (target, property, receiver) {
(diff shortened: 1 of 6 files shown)


src/JavascriptVM/JavascriptVM.tsx
```diff
@@ -102,24 +102,24 @@ export const VMProvider: FunctionComponent = ({ children }) => {
...
+  //             power: power,
+  //           })
+  //         );
+  //         return originalImpl.apply(target, [channel, power]);
+  //       };
+  //     }
+
+  //     return Reflect.get(target, property, receiver);
+  //   },
+  // };
```
JamesWP, line 122: we can remove these lines if they are not required anymore

ido-ran: done

PR #267, opened 2024-02-05
ci: fix artifact generation path in workflow

immutable-pro:
Previously, with CRA (Create React App), the build ouputs were put in `build`. Now, with Vite, we changed this to `dist`. Fixing this little error.


owen26 (approved):
nice!

PR #278, opened 2024-02-12
upgrade major runtime libraries

owen26:
Major changes including:

- react from 16 to 18, upgrade initialisation component and update explicit `children` prop
- blockly from 6 to 10, make a lot of breaking API changes compatible
- redux toolkit from 1.7 to 2.1
- js-interpreter from 2 to 5
- react-router-dom from 5 to 6, update routing setup with v6 breaking changes
- fix all lint errors
- remove google login dependencies and remove user/auth components as they are deprecated


owen26:
manually tested, and would be great if someone else can also double check and do a quick manual testing before vs after the upgrade

src/BlocklyInterface/blocklySlice.ts
```diff
@@ -138,7 +138,9 @@ export const saveBlocklyState = (state: RootState) => {
   try {
     const serializedState = JSON.stringify(state.blockly);
     localStorage.setItem(localStorageKey, serializedState);
-  } catch {}
+  } catch {
+    /* empty */
+  }
```
immutable-pro, line 143: Note: I know this is pre-existent code :). 

`stringify` will only throw if it finds a circular dependency. And `setItem` will only throw if the storage quota is exceeded.  In any of the cases, something fishy might be happening with the app if this throws. So let's not consume it. 

```suggestion
  } catch (err) {
    console.error(err)
  }
```

Or, **even better**, let's not catch it and make the app nicely crash.

immutable-pro (approved):
Thank you, @owen26, for the update and maintenance. 
I have reviewed the code and tested it locally. 

Unless the changes were very inter-dependent, I think smaller PRs will facilitate a bit the reasoning about the changes, when dealing with so many libraries for different use cases. But it looks clean and everything seems to be working, as far as I understand the app.

src/BlocklyInterface/blocklySlice.ts
owen26: yup that make sense, I was just did some quick lint fixing but didn't put too much thinking around the implementation

owen26:
Thanks @immutable-pro ! I promise this will be the last big PR I make 😂  So next time it'll be more reviewer friendly

PR #280, opened 2024-02-16
feat: setup semantic-release and bump node type to v20 align with nvm setup

owen26:
This PR mainly setup semantic-release so that each release will be versioned and change log will be more clear. For details please referring to the README file update.


immutable-pro (approved):
LGTM! Great idea!

PR #282, opened 2024-02-19
feat: new design layout

owen26:
This PR implements Kobot 2024 new design layout.

Highlights:

- Site overall structure implemented
- Header with navigation items, logo and feedback link implemented
- The header implementation kept responsiveness in mind, but it's not fully compatible with mobile devices since we don't really have design for that
- There is no Footer for the new design
- SVGR dynamic SVG loading module imported, this is useful for overriding SVG colors with CSS


src/ui/KobotThemeProvider.tsx
```diff
@@ -0,0 +1,9 @@
+import { FC, ReactNode } from "react";
+import { ThemeProvider } from "styled-components";
+import { theme } from "./theme";
+
+export const KobotThemeProvider: FC<{ children: ReactNode }> = ({
+  children,
+}) => {
+  return <ThemeProvider theme={theme}>{children}</ThemeProvider>;
+};
```
immutable-pro, line 9: Neat! 💅

immutable-pro (approved):
Great work, @owen26!
I checked out the code and tested it and seems to be working nicely.

PR #284, opened 2024-02-22
chore: disable PR comment on release success

owen26:
It feels more like spams generating new comment for every single PRs involved in a release. Disabling this behavior based on the docs here: https://github.com/semantic-release/github?tab=readme-ov-file#options


immutable-pro (approved):
Thank you!

PR #287, opened 2024-03-18
Initial components/changes as part of implementing Page 2 (Lessons)

yono10:
### Changes:

* Changes to Header, Button, and Typography to better align with design.
* Implement PageHeading component, and add use cases in the Sandbox and Lessons pages.
* Add usage of PageHeading to the Resources page too.
* Update the Components README accordingly.

#### Sandbox:

image

image

image

#### Page 2 / Lessons:

On ASUS Zenbook:

image

On iPad Pro (landscape):

image

On iPad Pro (portrait):

image


src/ui/PageHeading/PageHeading.tsx
```diff
@@ -0,0 +1,77 @@
+import { forwardRef } from "react";
```
owen26, line 1: might be better renaming "Heading" to something about "Header" to make it more UI component like

src/ui/PageHeading/PageHeading.tsx
```diff
@@ -0,0 +1,77 @@
...
+import shapesPink2Url from "./shapes-pink-2.svg";
+import shapesBlue1Url from "./shapes-blue-1.svg";
+import { H1, Subheading1, P } from "../../ui/Typography";
+
+type BackgroundType = "pink-1" | "pink-2" | "blue-1";
+
+const Wrap = styled.div<{
+  $backgroundType: BackgroundType;
+}>`
+  padding: 80px 40% 80px 11%;
```
owen26, line 13: if we want to keep right side of the panel empty, best way is to have a row in flexbox layout divided into 2 columns, with the right column stretch in portion by specifying `flex-grow` and `flex-shrink`

if we don't want that, we can let text flow to the right when device screen is smaller

src/ui/Typography.tsx
```diff
@@ -24,9 +24,14 @@ export const H4 = styled.h4`
 
 export const Subheading1 = styled.h5`
   font-size: var(--font-size-medium);
+  font-family: var(--font-heading-family);
```
owen26, line 27: From what I can see in Figma spec This is Gesta Bold 20px (labeled as pt but should be px)

Netlify demo site might styled it differently but the Figma one looks more like the intended design

src/ui/Typography.tsx
```diff
@@ -24,9 +24,14 @@ export const H4 = styled.h4`
...
 `;
 
 export const Subheading2 = styled.h6`
   font-size: var(--font-size-default);
   font-weight: var(--font-weight-bold);
 `;
+
+export const P = styled.p`
+  font-size: var(--font-size-default);
```
owen26, line 36: if it's already the default size we can have just a plain wrapper component `const P = styled.p```

src/ui/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,102 @@
...
+`;
+
+type Props = {
+  markerText: string;
+  heading: string;
+  description: string;
+  backgroundType?: BackgroundType;
+};
+
+export const PageHeader = forwardRef<HTMLDivElement, Props>(
```
owen26, line 87: Do we really need forward the ref here? I can't see it's been used anywhere.

src/ui/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,102 @@
...
+  margin: 0;
+`;
+
+const Heading = styled(H1)`
+  margin: 0;
+`;
+
+const Paragraph = styled(P)`
+  padding: 0 12% 0px 0px;
+  line-height: 19.2px;
```
owen26, line 76: unless there is a valid reason, normally we don't recommend have fraction of units, even if it appears in designer's file that normally are spcified unintentionally

src/ui/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const BlockHeader = styled.div`
+  flex-grow: 60;
+  flex-basis: 0;
+`;
+
+const BlockSpacer = styled.div`
+  flex-grow: 40;
+  flex-basis: 0;
+`;
```
owen26, line 53: normally it's good practice to use simplified ratio in flex grow/shrink, so in this case header gets grow:3, spacer gets grow:2

owen26:
Looking great, thanks for putting it together!

src/components/Header/Header.tsx
```diff
@@ -9,6 +9,8 @@ const StyledHeader = styled.header`
   align-items: stretch;
   background-color: var(--color-blue);
   flex: 0 0 56px;
+  box-shadow: 0px 2px 10px 0px rgba(7, 77, 118, 0.2);
+  z-index: 999;
```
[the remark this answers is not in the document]
yono10: OK yes can agree to that system. However I'd also suggest using 90 as the topmost layer (safe to say we won't have more than 10 layers for this site), as otherwise some arbitrary 30 doesn't strike immediately as being a top one.

src/ui/PageHeader/PageHeader.tsx
yono10: Those ones in the variables.css also do though, such as lineheight-text and lineheight-heading

src/ui/PageHeader/PageHeader.tsx
yono10: It was copy/paste job 🙂  - Although, I don't see the one in Button is used by any caller either?

src/ui/Typography.tsx
yono10: Subheading2 is, but Subheading1 is not though if you check carefully it's definitely a monospace, both in the Netlify and the Figma in fact.

src/components/Header/Header.tsx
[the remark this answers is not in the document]
owen26: absolutely, I would agree with any system as long as it's manageable

src/ui/PageHeader/PageHeader.tsx
owen26: `<Button />` is a bit different. Forward ref proxy the raw DOM `button` reference so that the consumer of `<Button />` component have access to it. This is useful and commonly a good practice to wrap interactive DOM element as React component so that callers can decide to manipulate with the raw element's click, focus, hover, readability events etc. Even though it's not yet used in our project.

Same would apply to wrap other interactive DOM elements like input, textarea, canvas etc.

src/ui/PageHeader/PageHeader.tsx
owen26: Probably because it's copy pasted from the demo site. It's my bad not double check those pasted code carefully :) Let's fix both by removing the fractions

src/ui/PageHeader/PageHeader.tsx
yono10: OK yes thanks for that explanation, I can see why it could be useful. I just thought I didn't see it being used currently. But of course if that's what its design is meant to be and just no use cases yet 👍

src/ui/PageHeader/PageHeader.tsx
yono10: So lineheight-text and lineheight-heading currently are 1.4 and 1.2, what should I change them to though? Or is that okay because it's not px rather?

owen26: Oh I see, sorry for not making to clear in my last comment, line-height without unit suffix is valid that indicating `1.2` times more than the original line-height. Another valid usecase for fractions is for example `1.234rem`, which is also indicating a ratio to original measurement.

However a fraction in absolute units like `px` would not be a common practice.

src/ui/Typography.tsx
yono10: I have updated these, hopefully the subheadings situation make more sense now 🙂

src/ui/PageHeader/PageHeader.tsx
yono10: Yes I suspected that then. Anyway I've updated the 19.2px to 19 now

owen26 (approved):
LGTM, thanks for putting it all together!

PR #288, opened 2024-03-27
Adding LessonCard as part of implementing Page 2 (Lessons)

yono10:
### Changes:
* Implement LessonCard component, and add a use case to implement the Lessons page further.
* Update the Components README accordingly.

Notes:
* LessonCard component is currently missing only one more thing: the counter (for number of challenges).
* Lesson Page is now only missing the InformationCard for "Guidance and resources" as the 6th card in the grid.

### Screenshots

On ASUS Zenbook:

image
image

On iPad Pro (landscape):

image

On iPad Pro (portrait):

image


owen26:
image
would be nicer here to have button always align to the bottom of the block (achieved by having a transparent div between description and button grow to fill the gap)

src/ui/LessonCard/LessonCard.tsx
```diff
@@ -0,0 +1,170 @@
...
+  align-items: center;
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: cover;
+  background-color: var(--color-blue-dark);
+
+  ${(props) =>
+    props.$mascotType === "lesson-1" &&
+    `
+    background-image: url(${shapesBgLesson1Url});
```
owen26, line 47: I think instead of having a `mascotType` and transplate it to a image url, it might be more straightforward to have image url as an input prop directly

```tsx
background-image: ${props => `url(${bgUrl})`};

```

So basically it's controlled from the parent component.

src/views/Lessons/Lessons.tsx
```diff
@@ -19,7 +26,48 @@ export const Lessons: FC = () => {
...
+          call2actionDest="../resources"
+        />
+        <LessonCard
+          mascotType="lesson-5"
+          subtitle="Lesson 5"
+          title="Against the clock"
+          description="Dolores et rerum hic quos sed at. Eos magnam laboriosam est repellat non. Dolores nihil suscipit quod voluptate."
+          call2actionText="View Challenges"
+          call2actionDest="../resources"
+        />
```
owen26, line 69: nit: would be good to have a json array object

```ts
const lessons = [
  {
    bgUrl,
    subtitle,
    title,
    description,
    call2actionText,
    call3actionDest,
  },
  ...
]

return <>
  ...
  {lessons.map((lesson, key) => <Lesson key={key} {...lesson} />)}
</>
```
so it can seamlessly migrate to a data persistent layer later.

However this is nit. feel free to address it in a future time.

owen26:
I think overall it's looking great and already handled responsiveness pretty well. Some minor details on edging cases I can might drop some extra comments but generally speaking I'm perfectly happy to merge it as it is.

owen26 (approved):
approved, feel free to either addrssing comment or just merge

PR #289, opened 2024-03-27
chore: adding e2e tests for home screen

owen26:
...and also fixes CI config

main changes:

- set playwright tests to only test on the Chromium browser so it's easier at an early stage building up test cases
- fix github workflow complaining about outdated node version
- fix home screen component throw warning with non-opaque styled component property name
- add home screen e2e test cases for playwright 
- configure dev server for playwright test


yono10 (approved):
Adding tests, should be good stuff 😄

Issue #272, opened 2024-02-08
Research which components we need for Page 4


cmanolescu:
[Page 4](https://www.figma.com/proto/QwuKynkYxQ67gwjZKrSubU/Kobot-UI-Ideas?node-id=662-18425)
[Components](https://lucky-smakager-43f6a1.netlify.app/#components-challenge-cards)

The outcome of this story should be a list of all components needed for this page

aram-cinnamon:
@cmanolescu I still can't assign myself...

PR #292, opened 2024-04-15
add resources page

aram-cinnamon:
https://github.com/FRUK-Simulator/Simulator/issues/272

Add secondary/tertiary buttons, progress div to lesson/resources card
Add lesson/resources card to resources page

Styled according to
https://lucky-smakager-43f6a1.netlify.app/#components-resource-cards
https://lucky-smakager-43f6a1.netlify.app/#forms-buttons-secondary
https://lucky-smakager-43f6a1.netlify.app/#forms-buttons-tertiary


src/components/LessonCard/LessonCard.tsx
```diff
@@ -108,63 +76,192 @@ const Paragraph = styled(P)`
...
-        <ImageMascot src={mascotUrl} />
+        <DivMascot>
+          {selectMascot(mascotType, isResourcesCard, progress)}
+        </DivMascot>
+        {!isResourcesCard && (
+          <DivProgress>
+            <img src={urlGearSvg} />
+            <span>{progress}/3</span>
+          </DivProgress>
+        )}
```
owen26, line 205: I would advice not to have a boolean or enum input prop to distinguish different implementation of a sharable component. It's against the principle of sharable component where it should have not kept knowledge of consumer specific implementations.

In this specific case, progress widget is only needed for resources card. Instead of having a `isResourcesCard` prop, we can replace it with a `progress` prop

```tsx
type UnderlyingCardProps = {
    ...
    progress: [number, number]
    ...
}

...
{!progress && (
    <DivProgress>
    <img src={urlGearSvg} />
    <span>{progress[0]}/{progress[1]}</span>
    </DivProgress>
)}

```

Then it's no longer about resouces card, but progress data, which is generic enough to be sharable for any component that needs a progress widget.

There is also a more flexible alternative approach:

We can have an empty slot injected in underlaying card and leave it to the consumer to decide what to render there.

```tsx
...
</DivMascot>
{children}
</BlockMascot>
...
```

Then the implementation of progress widget can be handed over as ResourcesCard's responsibility

```tsx
export const ResourcesCard = () => {
    return <UnderlayingCard {...}>
      <DivProgress>
        <img src={...} />
        <span>?/?</span>
        </DivProgress>
    </UnderlyingCard>
};
```

src/ui/Button.tsx
```diff
@@ -12,25 +14,26 @@ const StyledButton = styled.button<{
   justify-content: center;
   align-items: center;
   display: inline-flex;
+  margin: 0 0 var(--spacing-medium) 0;
```
owen26, line 17: I think margin should be specified by the parent container, instead of a built-in attribute to Button itself.

src/ui/Button.tsx
```diff
@@ -12,25 +14,26 @@ const StyledButton = styled.button<{
...
   position: relative;
   text-decoration: none;
-  box-shadow: 5px 5px 0 0 #000;
   font-family: var(--font-family-heading);
   white-space: nowrap;
 
   ${(props) =>
     props.$type === "primary" &&
     `
+    box-shadow: 5px 5px 0 0 #000;
```
owen26, line 35: let's use standard 6 digits color code `#000000` instead of shorthand form

src/ui/Button.tsx
```diff
@@ -70,6 +109,8 @@ export const Button = forwardRef<HTMLButtonElement, Props>(
         onClick={to ? () => navigate(to) : onClick}
       >
         {children}
+        {type === "secondary" && <IconDownload src={iconDownload} />}
+        {type === "tertiary" && <IconGo src={iconGo} />}
```
owen26, line 113: Adding icon to the content of the button should not be based on the type of the button as this component supposed to be shared for not only current usecases but all possible future use cases. We don't know if all future secondary buttons will all get a download icon.

src/components/LessonCard/LessonCard.tsx
```diff
@@ -108,63 +76,192 @@ const Paragraph = styled(P)`
...
-type Props = {
+const selectMascot = (mascotType: MascotType, isResourcesCard: boolean, progress: Progress) => {
+  const hideHead = isResourcesCard ? false : (progress <= 2);
+  const hideBody = isResourcesCard ? false : (progress <= 1);
+  const hideLegs = isResourcesCard ? false : (progress <= 0);
+
+  switch (mascotType) {
+    case 1:
+      return (
+        <svg data-progress={progress} xmlns="http://www.w3.org/2000/svg" id="Robot-1" viewBox="0 0 225.53 341.04">
```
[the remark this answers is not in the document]
yono10: I'm not sure I would be comfortable with burying logic/conditional inside an svg. An svg should just be plain static. Domain logic should be in the code.

In terms of the LessonCard owning that logic though - it was intentional by design in fact. So passing in an arbitrary url would mean the usage site will have to own that business logic. Which can still be fine, as long as we create some helper to facilitate that, so that Lessons/Resources/etc pages will not have to deal with sourcing those assets again themselves. 

Perhaps we should create a non UI component like AssetManager that loads all the assets, and then places like the LessonCard etc could just import that component, and then reference what they require (according to domain logic they should own and encapsulate). For example, LessonCard will reference `AssetManager.Shapes.LessonBackground1` when it needs the one for lesson 1, but it can remain the place to manage that logic, not on the usage site. Hope that makes sense.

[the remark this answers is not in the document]
yono10: Similarly, though separately, also on the logic around showing additional graphics. I'm not sure it's a good idea to have those optional as part of the graphics themselves. Again those are business logic that should live on the higher up layer, the UI rendering side. 

So with my original vision of the LessonCard, this is actually going to be just another parameter/config that can be enabled to optionally show the ProgressIndicator component, below the mascot figure, using Flex layout to manage that layout logic.

src/components/LessonCard/LessonCard.tsx
```diff
@@ -108,63 +76,192 @@ const Paragraph = styled(P)`
...
+        </svg>
+      );
+  }
+}
+
+type MascotType = 1|2|3|4|5;
+type Progress = 0|1|2|3;
```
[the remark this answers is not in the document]
yono10: (this comment shouldn't be read on its own, it ties back to what I described in the overall comment section first)

So on the progress indicator, I had thought about this being a standalone component in fact, per the audit done at https://github.com/FRUK-Simulator/Simulator/blob/main/src/components/README.md, see "ChallengesProgressIndicator". As this is also used (with variation showing total count, but therefore can just be made configurable) on the Challenges pages too, not just in Lessons page.

src/components/LessonCard/LessonCard.tsx
```diff
@@ -108,63 +76,192 @@ const Paragraph = styled(P)`
...
-      mascotUrl = shapesMascotLesson1Url;
-      break;
-  }
   return (
     <Wrap>
+
       <BlockMascot $mascotType={mascotType}>
-        <ImageMascot src={mascotUrl} />
+        <DivMascot>
+          {selectMascot(mascotType, isResourcesCard, progress)}
```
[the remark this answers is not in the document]
yono10: Related to my reply to the comment at https://github.com/FRUK-Simulator/Simulator/pull/292/files#r1579807948, this was actually intentional, so that the business logic of what assets make up "mascot" do not have to be exposed to the places that use the LessonCard. In terms of the asset lookup being done inside the component implementation if done the way as is, I think that can be mitigated by my proposal for just introducing an "AssetManager" component per my reply in there.

src/components/LessonCard/LessonCard.tsx
```diff
@@ -108,63 +76,192 @@ const Paragraph = styled(P)`
...
 `;
 
+const Hr = styled.hr`
+  margin: var(--spacing-x-large) 0 var(--spacing-x-large) 0;
+`;
+
 const StyledButton = styled(Button)``;
 
-type Props = {
+const selectMascot = (mascotType: MascotType, isResourcesCard: boolean, progress: Progress) => {
```
yono10, line 85: This again ties back to my explanation in the overall comment - I wouldn't call this specifically as "isResourcesCard", but to be about what specific pieces that are being configurable, eg. show title or not, show progress indicator, show which supplementary buttons, etc.

src/components/LessonCard/LessonCard.tsx
yono10: I appreciate that @owen26 comment here is a feedback based on continuing down this path of "UnderlyingCard", but see my comment about the `isResourcesCard`. I just don't think it is appropriate or necessary to introduce a distinct variant of this LessonCard, when as I originally envisioned it, it's simply a LessonCard, it's just it has optional bits that can be turned on/off.

src/views/Resources/Resources.tsx
```diff
@@ -1,7 +1,7 @@
 import styled from "styled-components";
 import { FC } from "react";
 import { PageHeader } from "../../components/PageHeader/PageHeader";
-import { UnderConstruction } from "../../components/UnderConstruction/UnderConstruction";
+import { ResourcesCard } from "../../components/LessonCard/LessonCard";
```
yono10, line 4: Initially, I was worried that I thought you had created a completely separate ResourcesCard! Until I saw that this is just being aliased here. But therefore on that note, I don't think it is a good practice to do that. If it's a LessonCard component, then it should still be named a LessonCard as far as the import.

yono10 (commented):
So looks like I failed to communicate how I envisioned the LessonCard component to be (due to me leaving it partly implemented too before I had to leave), especially in how it was going to be reused elsewhere beyond the Lessons page including the Resources page therefore. 

Some of my inline comments here therefore reflect my feedback that are based on how the changes would be incompatible to how the component was intended to be designed. 

To give details of how I envisioned the design for the LessonCard component:
* It is designed to be reusable, based on known usages in Lessons and Resources page.
* It knows how to render a Lesson Card with the appropriate mascot background & figure depending on lesson number.
* It has a basic layout that include the mascot, the title, subtitle, and description.
* It can be configured to optionally show a Call To Action button (rendered as primary).
* It can be configured to optionally show a progress indicator below the mascot figure.
* It can be configured to optionally show two more supplementary buttons, rendered as secondary and tertiary buttons, as additional rows after the Call To Action button (ie. this is for the Download Resources and Go To Lesson in the Resources page, for example).

Additionally - as part of reviewing this PR, I would also propose that we introduce a non-UI component say called `AssetManager` that will be the central place to load those assets and then provide those resources as namespace-based identifiers that the UI components can then reference easily, while therefore not needed to be doing the asset lookup within themselves. If we're worried about perf - we can also perhaps make those identifiers "lazy" - such as only on being referenced, that the actual lookup happens, I'd think.

owen26:
> as part of reviewing this PR, I would also propose that we introduce a non-UI component say called AssetManager that will be the central place to load those assets

I've done something as part of my challenges page implementation, not sure if you are referring to something similar.

https://github.com/FRUK-Simulator/Simulator/blob/main/src/core/data-loader.ts

yono10:
> > as part of reviewing this PR, I would also propose that we introduce a non-UI component say called AssetManager that will be the central place to load those assets
> 
> I've done something as part of my challenges page implementation, not sure if you are referring to something similar.
> 
> https://github.com/FRUK-Simulator/Simulator/blob/main/src/core/data-loader.ts

I think not exactly (unless I missed it somewhere), as in there I see ChallengeCard still accepting a plain url as a parameter - which means it doesn't own that logic (of what makes up an icon image for a Challenge), and that any usage site can just pass in any image into it. 

What I was thinking/proposing, would be somethign like (total pseudocode!):

```
ChallengeCard {
  let iconSvg = when (challengeType) {
    is A -> AssetManager.Shapes.Icons.ChallengeA,
    is B -> AssetManager.Shapes.Icons.ChallengeB,
    is C -> AssetManager.Shapes.Icons.ChallengeC,
  } 
  ...
}

AssetManager {
  Shapes {
    Icons {
      ChallengeA = ... <svg lookup/import>,
      ChallengeB = ... <svg lookup/import>,
      ...
    }

    Backgrounds {
      ...
    }
  },
  ...
}
```

owen26:
> I think not exactly...

Right.. Cuz I see those cover images as part of lessons/challenges/resources data structure no difference to other fields for example titles and descriptions. If we later add more lessons to the dataset then it's likely there will be new cover images attached to them each representing some nature of the lesson itself. This is unlike generic reusable images like background patterns for example.

So I would argue asset lookup based on type is not necessary because each image is unique and represents only the specific lesson, so it makes more sense to be part of the lesson object.

But ofc I'm open to further discussion :)

yono10:
> > I think not exactly...
> 
> Right.. Cuz I see those cover images as part of lessons/challenges/resources data structure no difference to other fields for example titles and descriptions. If we later add more lessons to the dataset then it's likely there will be new cover images attached to them each representing some nature of the lesson itself. This is unlike generic reusable images like background patterns for example.
> 
> So I would argue asset lookup based on type is not necessary because each image is unique and represents only the specific lesson, so it makes more sense to be part of the lesson object.
> 
> But ofc I'm open to further discussion :)

OK yes I had missed that those images will in fact need to be data driven, so will need to come from database indeed.
Basically my concern is that the components shouldn't become just generic containers when they have been decidedly called "Lesson Card", "Challenge Card" etc, so they should be by design quite self-sufficient in terms of implementing those characteristics. But yes where the pieces of information are actually those that are data driven, then sure they need to be passed in. (still alternatively though, we could also let them consult something like AssetManager.getLessonBackground() passing in lesson id, if there the urls are standardized into a certain known location, as opposed to being made as part of the lesson data).

Regardless, for non data driven ones though, I do think they should be part of the component implementation details, and not let the callers worry about passing them in.

src/components/LessonCard/LessonCard.tsx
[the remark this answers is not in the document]
yono10: > I'm not sure I would be comfortable with burying logic/conditional inside an svg. An svg should just be plain static. Business logic should be in the code.
> 
> In terms of the LessonCard owning that logic though - it was intentional by design in fact. So passing in an arbitrary url would mean the usage site will have to own that business logic (and it may also be nothing to do with lessons!). Which can still be fine, as long as we create some helper to facilitate that, so that Lessons/Resources/etc pages will not have to deal with sourcing those assets again themselves.
> 
> Perhaps we should create a non UI component like AssetManager that loads all the assets, and then places like the LessonCard etc could just import that component, and then reference what they require (according to domain logic they should own and encapsulate). For example, LessonCard will reference `AssetManager.Shapes.LessonBackground1` when it needs the one for lesson 1, but it can remain the place to manage that logic, not on the usage site. Hope that makes sense.

On the logic/conditional inside svgs though, I'm really not sure this is the right way to go. Thoughts @owen26 ?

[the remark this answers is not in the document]
owen26: my opinion is the same as my original comment, this SVG should be imported as an asset URL instead of a React component/node. This also aligns with my view on all images so far in this project. So that those assets are made compatible with external data persistence services such as a CMS. And they can be in other form not only SVG but JPEG and PNG.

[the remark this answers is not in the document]
yono10: OK cool thanks. And just to be clear, my concerns were not just on the inline SVG, but also the use of the variables inside the SVG.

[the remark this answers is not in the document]
owen26: Yes I agree. Inline SVG with variables is rarely used in practice. To be fair it can be powerful in specific use cases. But here however as I mentioned those SVG better be served as external assets rather than inline into component, so no variables involved.

yono10:
> > > I think not exactly...
> > 
> > 
> > Right.. Cuz I see those cover images as part of lessons/challenges/resources data structure no difference to other fields for example titles and descriptions. If we later add more lessons to the dataset then it's likely there will be new cover images attached to them each representing some nature of the lesson itself. This is unlike generic reusable images like background patterns for example.
> > So I would argue asset lookup based on type is not necessary because each image is unique and represents only the specific lesson, so it makes more sense to be part of the lesson object.
> > But ofc I'm open to further discussion :)
> 
> OK yes I had missed that those images will in fact need to be data driven, so will need to come from database indeed. Basically my concern is that the components shouldn't become just generic containers when they have been decidedly called "Lesson Card", "Challenge Card" etc, so they should be by design quite self-sufficient in terms of implementing those characteristics. But yes where the pieces of information are actually those that are data driven, then sure they need to be passed in. (still alternatively though, we could also let them consult something like AssetManager.getLessonBackground() passing in lesson id, if there the urls are standardized into a certain known location, as opposed to being made as part of the lesson data).
> 
> Regardless, for non data driven ones though, I do think they should be part of the component implementation details, and not let the callers worry about passing them in.

Obviously much time has elapsed since, and some side discussions has happened too. So just to summarise here, that my suggestion of AssetManager is no longer very essential at this point (even in terms of for the statically loaded assets, ie. non data driven - I think doing that would still be ideal, but perhaps we can improve that later too). For the data driven ones, as previously advised by @owen26 , let's make the component accept the url in, so in the caller site once they have loaded it from the data source, they can pass in the url accordingly.

On the other hand - @aram-cinnamon In the interest of moving this PR forward though, could you look at addressing some of the other outstanding topics that have been raised? @owen26 might want to add, but from what I can tell here are the outstanding items:
* Could you please extract SVG assets out of the components, and using imports instead.
* Let's not use variables inside the SVGs, especially for toggling on/off different parts of the SVG when they can/should just be split into multiple assets imported in separately.
* The buttons on the LessonCard will be required to perform different navigations. Originally I had this as "call2actionText/call2actionDest" pair (with the thinking that we'll just add more such pairs for the secondary and tertiary ones), but that was removed by one of @owen26's changes in favour of the "lessonId". I still think we need that more generic flexibility, because now you have more buttons, they will need to perform different navigations, so I think we need to accept those in as arguments, like my original design. (minor note: call2action stands for "Call to action" 🙂 )

owen26:
feel free to re-request me review when there are updates :)

src/components/LessonCard/LessonCard.tsx
```diff
@@ -1,80 +1,77 @@
...
-import shapesBgLesson2Url from "./shapes-bg-lesson2.svg";
-import shapesBgLesson3Url from "./shapes-bg-lesson3.svg";
-import shapesBgLesson4Url from "./shapes-bg-lesson4.svg";
-import shapesBgLesson5Url from "./shapes-bg-lesson5.svg";
-import shapesMascotLesson1Url from "./shapes-mascot-lesson1.svg";
-import shapesMascotLesson2Url from "./shapes-mascot-lesson2.svg";
-import shapesMascotLesson3Url from "./shapes-mascot-lesson3.svg";
-import shapesMascotLesson4Url from "./shapes-mascot-lesson4.svg";
-import shapesMascotLesson5Url from "./shapes-mascot-lesson5.svg";
+import urlLesson1MascotSvg from "./lesson-1-mascot.svg";
```
yono10, line 3: I think as discussed with @owen26 previously, let's have these urls passed into this component (as they will be as loaded from the data source) instead of getting them baked in here.

yono10: As this was pre-existing pattern as I also originally did it - happy for that to be done in a separate PR if you prefer.

src/components/LessonCard/LessonCard.tsx
```diff
@@ -108,61 +108,147 @@ const Paragraph = styled(P)`
...
+  margin: var(--spacing-x-large) 0 var(--spacing-x-large) 0;
+`;
+
 const StyledButton = styled(Button)``;
 
+const selectMascot = (mascotType: MascotType, progress?: Progress) => {
+  let maskHead = false;
+  let maskBody = false;
+  let maskLegs = false;
+  if (progress) {
```
yono10, line 121: This has been confusing to me - what is this masking logic actually for? Is this about rendering the mascot only half way according to the progress? 

If so - while the implementation is neat, but I'm not sure where that requirement comes from as I can't see that anywhere on the design guides we have been given? I don't see it in the figma nor the prototype. Have I missed it somewhere - could you point to that?

src/components/LessonCard/LessonCard.tsx
```diff
@@ -108,61 +108,147 @@ const Paragraph = styled(P)`
...
+  mascotType: MascotType;
+  progress?: Progress;
   subtitle?: string;
   title: string;
   description: string;
-  call2actionText?: string;
-  mascotType: MascotType;
+  buttonTextPrimary?: string;
+  buttonDestPrimary?: string;
+  buttonTextSecondaryDownload?: string;
```
yono10, line 190: Can we not call this specific to "Download". Calling it secondary to be generic seems sufficient.

src/ui/Button.tsx
```diff
@@ -1,8 +1,10 @@
 import { forwardRef, ReactNode } from "react";
 import { useNavigate } from "react-router-dom";
 import styled from "styled-components";
+import iconDownload from "./icon-download.svg";
+import iconGo from "./icon-go.svg";
 
-type ButtonType = "primary" | "primary-blue" | "secondary" | "tertiary";
+type ButtonType = "primary" | "primary-blue" | "secondary" | "secondary-download" | "tertiary";
```
yono10, line 7: Why is it that we need a different "secondary-download" type? 

Looking at the design for Resources page, those download buttons to me look like the exact style for the button called "secondary" in the design system anyway?

src/views/Lessons/Lessons.tsx
```diff
@@ -28,44 +28,49 @@ export const Lessons: FC = () => {
...
-          lessonId="1"
-          mascotType="lesson-1"
+          mascotType={1}
+          progress={{ completed: 0, total: 3 }}
           subtitle="Lesson 1"
           title="Motors"
           description="Dolores et rerum hic quos sed at. Eos magnam laboriosam est repellat non. Dolores nihil suscipit quod voluptate."
-          call2actionText="View Challenges"
+          buttonTextPrimary="View Challenges"
+          buttonDestPrimary="#"
```
yono10, line 37: Why '#'? Should be 1, 2, etc?

yono10 (commented):
Had another pass - thanks for the changes. Left a couple comments/questions. 

My main reservation is around the mascot masking logic, as I'm not sure where that requirement comes from, if you can clarify. Thanks.

src/components/LessonCard/LessonCard.tsx
aram-cinnamon: Hm yes, let's do it in a separate PR. :)
I actually prefer to have the lesson mascots here, since they're used only here in the LessonCard. No need for other components to know about them.
I could be persuaded otherwise though. I don't feel that strongly about it.

src/components/LessonCard/LessonCard.tsx
aram-cinnamon: https://lucky-smakager-43f6a1.netlify.app/#components-lesson-cards
"The Lesson Cards use an inline SVG image with names zones of Head, Body and Legs, each SVG also has a white merged version of each part named head-bg, bogy-bg and legs-bg.
We target the visibility of these parts by passing a data attribute ( data-progress="0") of 1, 2 or 3."

This requirement is why I had originally put the SVG assets in LessonCard.tsx. That's honestly the cleanest implementation, without awkward `position`ing in the css.

src/components/LessonCard/LessonCard.tsx
aram-cinnamon: Hm yes, let's do it in a separate PR. :)
I actually prefer to have the lesson mascots here, since they're used only here in the LessonCard. No need for other components to know about them.
I could be persuaded otherwise though. I don't feel that strongly about it.

src/ui/Button.tsx
aram-cinnamon: "secondary" doesn't include the download icon; "secondary-download" does.

I added the "secondary-download" button type in response to a comment above: "We don't know if all future secondary buttons will all get a download icon."

If ever we need a "secondary" button that shouldn't have a download icon, we can use "secondary."
Meanwhile, the name "secondary-download" signals that it should only be used for buttons that let you download stuff.

src/components/LessonCard/LessonCard.tsx
aram-cinnamon: Please see the answer below

src/views/Lessons/Lessons.tsx
aram-cinnamon: Do we know yet what all the `Dest`s should be? (e.g. What's the resource link for Download Resources? I don't know.)

yono10: Yes, if you see the original version, that was working right? It was 1, 2, 3 etc, and clicking them will go to the challenge page that @owen26 created iirc?

src/components/LessonCard/LessonCard.tsx
yono10: Oh interesting, I had completely missed that! I am not sure why they didn't reflect that in the figma either!

Fair enough. Anyway with that in mind, I think your current implementation is cool though. I do think it's cleaner this way with the separate assets, and I can see the mask overlay logic clearly too. Thanks 👍

yono10 (commented):
Had another pass again, thanks @aram-cinnamon for clarifying some of the points I raised. 

So I guess now the only outstanding thing for me is about the "secondary-download" button type. Per my comment, I'd vote for making button accepts icon to be passed in, instead of baking in this specific "download type". @owen26 Maybe you can do the tie breaker 🙂 

(Also - I have marked as resolved some of the comments I was involved in and have since moved on, I have left those raised by @owen26 and remain unanswered, so you might want to check those directly with him)

owen26:
> Had another pass again, thanks @aram-cinnamon for clarifying some of the points I raised.
> 
> So I guess now the only outstanding thing for me is about the "secondary-download" button type. Per my comment, I'd vote for making button accepts icon to be passed in, instead of baking in this specific "download type". @owen26 Maybe you can do the tie breaker 🙂
> 
> (Also - I have marked as resolved some of the comments I was involved in and have since moved on, I have left those raised by @owen26 and remain unanswered, so you might want to check those directly with him)

Yes I would strongly advice not to implement `secondary-download`, but instead the icon should be part the label

```
<Button type="secondary">
  <DownloadIcon />
  Download
</Button
```

src/ui/Button.tsx
```diff
@@ -108,7 +101,6 @@ export const Button = forwardRef<HTMLButtonElement, Props>(
         onClick={to ? () => navigate(to) : onClick}
       >
         {children}
-        {type === "secondary-download" && <IconDownload src={iconDownload} />}
         {type === "tertiary" && <IconGo src={iconGo} />}
```
yono10, line 104: I suppose, similarly, "tertiary" is also being hardcoded to "Go" icon.  We should also do similar there, let caller passes that in.

src/views/Lessons/Lessons.tsx
```diff
@@ -43,7 +43,7 @@ export const Lessons: FC = () => {
           title="Distance sensors"
           description="Dolores et rerum hic quos sed at. Eos magnam laboriosam est repellat non. Dolores nihil suscipit quod voluptate."
           buttonTextPrimary="View Challenges"
-          buttonDestPrimary="#"
+          buttonDestPrimary="1"
```
yono10, line 46: I thought this should be 1, 2, 3, 4, 5.. but it was @owen26 who sets up those pages anyway so I'm not sure if it was placeholder or not? If it was real pages, then might want to use them?

yono10 (approved):
Thanks for the changes. I have left a comment/question or two. But happy for this to land first and get those updated as a follow up if need be.

src/views/Lessons/Lessons.tsx
owen26: for now I think making them all `1` is fine, we should have a follow up PR unify data loading machanism, then this `buttonDestPrimary` will be dynamically set to real lesson ID fetched from data source.

src/ui/Button.tsx
owen26: +1

aram-cinnamon: https://lucky-smakager-43f6a1.netlify.app/#forms-buttons-tertiary
The "Go" icon is very much part of the tertiary button. It's the only thing that marks the tertiary button as a button!

aram-cinnamon:
do you also have to approve @immutable-pro ? @cmanolescu ?

PR #293, opened 2024-04-30
ci: fix deployments by build outDir changed to `build` and re-enabling artifact generation

immutable-pro:
- **ci: changing output directory for prod builds to `build`**
- **ci: re-enabling artifact.zip generation**

### What's going on?

The machine where the website is deployed is periodically checking for the existence of an `artifact.zip` file, containing a `build` folder inside, with the required assets for the deployment.

A manual Github workflow has been added to trigger the release.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -7,7 +7,7 @@ module.exports = {
     "plugin:react-hooks/recommended",
     "plugin:@tanstack/eslint-plugin-query/recommended",
   ],
-  ignorePatterns: ["dist", ".eslintrc.cjs"],
+  ignorePatterns: ["dist", "build", ".eslintrc.cjs"],
   parser: "@typescript-eslint/parser",
   plugins: ["react-refresh"],
   rules: {
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,8 +24,21 @@ jobs:
         env:
           CI: true
 
+      - run: zip -r artifact.zip build
+
       - name: Create Release
         id: create_release
         run: npx semantic-release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload Release Asset for deployment scripts
+        id: upload-release-asset
+        uses: actions/upload-release-asset@v1
+        env:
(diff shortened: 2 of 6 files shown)


owen26 (commented):
Looking great, thanks for putting it all together.

Only question before approve, is that is this going to be deployed to the one and only environment for kobot? Which is essentially production?

We might want to double check with FIRST to see if it's fine to override the current legacy site if that's the case.

Long term speaking it would be great to have this CI deployed to dev or uat environment, and have a prod deploy manually triggered.

immutable-pro:
> Looking great, thanks for putting it all together.
> 
> Only question before approve, is that is this going to be deployed to the one and only environment for kobot? Which is essentially production?
> 
> We might want to double check with FIRST to see if it's fine to override the current legacy site if that's the case.
> 
> Long term speaking it would be great to have this CI deployed to dev or uat environment, and have a prod deploy manually triggered.

Yes, you are right. I thought about that too. It is a bit ...aggressive. 

Before merging, I was going to suggest a "plan B" approach that consists of:
- Use GitHub pages as a deployment preview. 
- Only generate `artifacts.zip` when a certain tag is created in GitHub, so we can control PROD deployments. 

For that purpose, I would need admin access of the repo, as I can't see how GitHub pages is set up today, if at all.

owen26:
> Yes, you are right. I thought about that too. It is a bit ...aggressive.
> 
> Before merging, I was going to suggest a "plan B" approach that consists of:
> 
> * Use GitHub pages as a deployment preview.
> * Only generate `artifacts.zip` when a certain tag is created in GitHub, so we can control PROD deployments.
> 
> For that purpose, I would need admin access of the repo, as I can't see how GitHub pages is set up today, if at all.

Yes this looks like a great plan. The current site is totally static so GitHub Pages would be a good fit for deployment. It might also has the extra benefit for not being blocked by our VPN cuz I think Netlify on the other hand is blocked.

For production deployment yes we can either generate artifact when a tag is generated. (aka. `feat` or `fix` commit involved)

Alternatively, we can also have the artifact build workflow targeting `on: workflow_dispatch` trigger on github, so we can a manual "Run workflow" button to make prod deployment.

